PR #9, opened 2022-06-22
feat: add namespace contract

varunsrin:


src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+
+import {FixedPointMathLib} from "../lib/solmate/src/utils/FixedPointMathLib.sol";
+
+error InsufficientFunds(); // The transaction does not have enough money to pay for this.
+error Unauthorized(); // The caller is not authorized to perform this action.
+
+error InvalidCommit(); // The commitment hash was not found
+error InvalidName(); // The username had invalid characters
+error InvalidTime(); // Time is too far in the future or past
+error IncorrectOwner(); // The username is not owned by the expected address
```
nichochar, line 16: When would we use this over `Unauthorized`, line 11 ? Opportunity to simplify and keep only one?

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+
+error Registered(); // The username is currently registered.
```
nichochar, line 18: Nit: /s/Registered/AlreadyRegistered

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+error NotRegisterable(); // The username has been registered and cannot be registered again.
+error Registerable(); // The username has never been registered.
+
+error Expired(); // The username is expired (renewable or biddable)
```
nichochar, line 22: Nit: names of errors don't specify that it's for usernames. `ExpiredUsername` would be more explicit

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+    /*//////////////////////////////////////////////////////////////
+                        RECOVERY STORAGE
+    //////////////////////////////////////////////////////////////*/
+
+    // Mapping from tokenId to recoveryAddress
+    mapping(uint256 => address) public recoveryOf;
+
+    // Mapping from tokenId to recovery timestamp in seconds, which is set to zero on cancellation
+    // or completion
+    mapping(uint256 => uint256) public recoveryClockOf;
```
nichochar, line 75: I'm not a gas golf expert, but curious if having a mapping of `(uint256 => recoveryObj)` where recoveryObj is a struct with recovery address, clock and recoveryDestination would be more efficient.

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+    //////////////////////////////////////////////////////////////*/
+    string public baseURI = "http://www.farcaster.xyz/";
+
+    uint256 public immutable gracePeriod = 30 days;
+
+    uint256 public immutable fee = 0.01 ether;
+
+    // The epoch timestamp of Jan 1 for each year starting from 2022
+    uint256[] internal _yearTimestamps = [
+        1640995200,
```
nichochar, line 92: These could do with comments for each of the years

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+     * @dev The commitment process prevents front-running of the username registration.
+     *
+     * @param commit the commitment hash to be persisted on-chain
+     */
+    function makeCommit(bytes32 commit) public {
+        stateOf[commit] = true;
+    }
+
+    /**
+     * @notice Mint a new username if a commitment was made previously and send it to the owner.
```
nichochar, line 169: You could add a comment here that it is expected that `makeCommit` was called before this runs, or it reverts

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+        if (msg.value < _currYearFee) revert InsufficientFunds();
+
+        if (!stateOf[commit]) revert InvalidCommit();
+        delete stateOf[commit];
+
+        uint256 tokenId = uint256(bytes32(username));
+        if (expiryOf[tokenId] != 0) revert NotRegisterable();
+
+        _mint(owner, tokenId);
+
```
nichochar, line 192: Should we emit an event here for registration? (or are we assuming the _mint one is enough)

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+        // Calculate the num of 8 hr periods since expiry as a fixed point signed decimal. The
+        // constant approximates fixed point division by 28,000 (num of seconds in 8 hours)
+        int256 periodsSD59x18 = int256(3.57142857e13 * (block.timestamp - auctionStartTimestamp));
+
+        // Optimization: precompute return values for the first few periods and the last one.
+
+        // Calculate the price by taking the 1000 ETH premium and discounting it by 10% for every
+        // period and adding to it the renewal fee for the current year.
+        uint256 price = uint256(1_000 ether).mulWadDown(
```
nichochar, line 268: Should `1_000 ether` be made a variable that the contract owner can update at a later time?

Otherwise, just for readability it could be made a named constant

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+    /**
+     * @notice Set a recovery address which can transfer the caller's username to a new address.
+     *
+     * @param recoveryAddress the recoveryAddress, which must not be the custodyAddress. It can be
+     *                        set to zero to disable the recovery functionality.
+     */
+    function setRecoveryAddress(uint256 tokenId, address recoveryAddress) external payable {
+        if (ownerOf(tokenId) != msg.sender) revert Unauthorized();
+
+        if (recoveryAddress == msg.sender) revert InvalidRecovery();
```
nichochar, line 353: Could also check for the null address as an invalid one here

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
...
+            if (
+                (charInt >= 1 && charInt <= 44) ||
+                (charInt >= 46 && charInt <= 47) ||
+                (charInt >= 58 && charInt <= 96) ||
+                charInt >= 123
+            ) {
+                return false;
+            }
+
+            // i can never overflow because length is guaranteed to be < 16
```
nichochar, line 571: That check doesn't happen in this function. Might be worth adding to prevent developer errors calling this without checking

src/Namespace.sol
varunsrin: updateable values and named constants require slightly more gas, and the former also adds more surface area. i'd prefer to just do a full contract upgrade and keep the value immutable.

src/Namespace.sol
varunsrin: i believe it is implicitly guaranteed -- length is calculated from .length called on a bytes16 value, so there is no way you should able to call this function and cause length to exceed 16

src/Namespace.sol
```diff
@@ -0,0 +1,578 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.14;
+
+import {ERC721} from "../lib/solmate/src/tokens/ERC721.sol";
+
+import {Owned} from "../lib/solmate/src/auth/Owned.sol";
+
+import {FixedPointMathLib} from "../lib/solmate/src/utils/FixedPointMathLib.sol";
+
+error InsufficientFunds(); // The transaction does not have enough money to pay for this.
```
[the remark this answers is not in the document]
varunsrin: what detail would you want to see in the string?

src/Namespace.sol
varunsrin: so the case here is that `bob` is renewing the name `@foo` that `alice` currently owns. bob wants to specify that is paying the fee for foo and for alice owning foo. 

if somehow alice sells the name to charlie or an admin reclaimed this name, bob doesn't want this transaction to complete. so if the name did change hands and it is now owned by charlie, we throw IncorrectOwner. 

Unauthorized is slightly misleading here - bob *is* authorized to pay to renew charlie's username `!@foo` its just that bob only wanted to pay it when alice owned it.

varunsrin: open to other terms, I agree that incorrect owner is not immediately obvious to the reader

PR #81, opened 2022-08-28
feat(IDRegistry): prevent registration and transfer to the zero address

varunsrin:


varunsrin:
on reflection, this change is unnecessary because our id ownership logic will prevent the zero address from ever owning > 1 id due to the constraints placed around id ownership.

PR #100, opened 2022-09-05
fix(BundleRegistry): invitee should be automatically attributed

varunsrin:
Previously, BundleRegistry's `trustedRegister` and `partialTrustedRegister` required passing in the invitee for the Invite event. However, this is only known after calling IDRegistry during the call and cannot be provided ahead of time. This change no longer requires the parameter to be provided and populates it automatically. 

There are two ways this could have been implemented: 

1. Modifying IDRegistry's `register` and `trustedRegister` to return the id
2. Query the IDRegistry's idOf function between calls to fetch the id 

Option 1 was rejected because it added a fixed ~150 gas cost to all register calls for all time, while Option 2 only added ~800 gas to calls made during the invitation period.  While Option 1 adds greater composability for future actions that might want to do things with the id immediately after registration, those use cases are less clear and so Option 2 was selected instead.


varunsrin:
cc @pfletcherhill

PR #104, opened 2022-09-06
chore: improvements to default configuration

varunsrin:
- enforce default line length of 120 for solidity, 100 otherwise
- tag the .sol files so they show up in github search correctly using gitattributes
- increase optimizer runs to prioritize run efficiency over deploy size efficiency
- reduce verbosity in ci to only show logs when tests fail
- reduce default fuzzer runs to 512 to run faster 

inspired by defaults in https://github.com/transmissions11/foundry-template


varunsrin:
reverting the optimizer improvement since it caused the NameRegistry to become ~3kb larger which made it undeployable

Issue #48, opened 2022-08-10
testing: add precommit hook to check if gas snapshots are up to date


varunsrin:
gas snapshots that are out of date cause the ci to fail, and it's very easy to forget to run this leading to a lot of wasted ci runs. 

add a pre-commit hook that runs `forge snapshot` on commit and ensures that it is up-to-date before allowing the commit to happen.

PR #118, opened 2022-09-09
feat: add precommit hook to check if gas snapshots are up to date

craigjson:
addresses #48 

Question: I noticed the CI is running "--match-contract Gas" which I have included here to keep it in line with the expected results. I see that we also have some tests that are specifically related to Gas usage. Is there a downside to running the snapshot for all tests?

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn lint-staged
+forge snapshot --check --match-contract Gas


varunsrin:
thanks for this PR! 

the main goal of forcing snapshots in the CI is to make it easy for a reviewer to know if the gas costs for something are going up.  if you see a snapshot diff, then it's a sign to dig in and investigate. 

the problem is that a snapshot diff happens whenever the cost of *running a test* changes even if it has no impact on the contract's actual logic. so if you turn them on for all tests, it becomes extremely noisy and is basically useless as a tool to quickly see if your contract is getting more expensive.

craigjson:
> thanks for this PR!
> 
> the main goal of forcing snapshots in the CI is to make it easy for a reviewer to know if the gas costs for something are going up. if you see a snapshot diff, then it's a sign to dig in and investigate.
> 
> the problem is that a snapshot diff happens whenever the cost of _running a test_ changes even if it has no impact on the contract's actual logic. so if you turn them on for all tests, it becomes extremely noisy and is basically useless as a tool to quickly see if your contract is getting more expensive.

ah I see, totally makes sense to me that you'd want the snapshots to be readable as possible. Leet me know if there's any changes you'd like made to this PR.

Issue #109, opened 2022-09-07
perf: NameRegistry gas costs can be reduced by increase optimizer runs


varunsrin:
Increase solidity optimizer runs to 50,000 or higher makes some gas improvements for register(), renew() and bid() which are commonly used methods. Unfortunately, this increases the size of the contract by ~3kb which makes it too large to deploy. 

Investigate if there are any reasonable means to reduce contract size, which could allow us to take advantage of these gas cost reductions.

varunsrin:
Implementing this is likely contingent on our choice of upgrade mechanism which isn't being done in this milestone. Removing the flag

varunsrin:
Closing since NameRegistry was deprecated

PR #172, opened 2023-02-24
feat: simple metadata packing gas optimization, saves 20k gas per registration

shotaronowhere:
## Motivation

Over-optimization can lead to complexity --- bad for security and auditing.

On the other hand, however, simple storage packing optimizations, especially when they reduce the gas of the most common user facing interactions eg. `register(...)`, can be more than worth it.

The metadata stored in mappings when registering and recovering fnames is wasteful. Especially if the contract is deployed on mainnet, best not to wast precious state.

Timestamps do not require 256 bits. 40 bits is sufficient for the next ~40,000 years. Even 32 bits is sufficient for ~80 years.

`(2^40-1677253973)/60/60/24/365 = 34812 years`

Addresses can be packed together with timestamps in a single slot (160 bit addresses, 40 bit timestamps << 256 bits in a single slot).

## Change Summary

Packed timestamp and address metadata in a simple yet gas saving optimization. Saves 20k gas per registration and recovery.

## Merge Checklist

## Additional Context

n.a.


src/NameRegistry.sol
```diff
@@ -223,44 +243,29 @@ contract NameRegistry is
...
     /**
-     * @notice Maps each uint256 representation of an fname to the destination address of the most
-     *         recent recovery attempt.
-     * @dev    Occupies slot 9, and the value is left dirty after a recovery to save gas and should
-     *         not be relied upon to check if there is an active recovery.
+     * @notice Maps each uint256 representation of an fname to recovery metadata
+     * @dev    Occupies slot 7
      */
-    mapping(uint256 => address) public recoveryDestinationOf;
+    mapping(uint256 => RecoveryMetadata) public recoveryInfo;
```
varunsrin, line 268: can you update this to `recoveryMetadataOf` to stick with the existing conventions for map names?

src/NameRegistry.sol
```diff
@@ -223,44 +243,29 @@ contract NameRegistry is
...
-
-    /**
-     * @notice Maps each uint256 representation of an fname to the timestamp of the recovery
-     *         attempt or zero if there is no active recovery.
-     * @dev    Occupies slot 8
+     * @notice Maps each uint256 representation of an fname to registration metadata
+     * @dev    Occupies slot 6
      */
-    mapping(uint256 => uint256) public recoveryClockOf;
+    mapping(uint256 => RegistrationMetadata) public registrationInfo;
```
varunsrin, line 262: can you update this to `registrationMetadataOf` to stick with the existing conventions for map names?

src/NameRegistry.sol
```diff
@@ -44,6 +44,26 @@ contract NameRegistry is
         address destination;
     }
 
+    /**
+     * @dev RegistrationMetadata struct represents the fname registration data.
+     * @param recoveryOf The address that can recover the fname.
+     * @param expiryOf The time at which fname expires.
+     */
+    struct RegistrationMetadata {
+        address recoveryOf;
```
varunsrin, line 53: the attributes can just be called recovery, expiryTs, recoveryDestination, recoveryTs since they are no longer map names

varunsrin (changes requested):
Looks great, thanks for the gas opts!

The slot changes means this will no longer be upgradeable on goerli, which is fine since we plan to redeploy at some point. But we may want to keep this as a branch for a few more weeks until we are confident we don't need to upgrade the existing contract. 

You'll also need to [sign your commits](https://github.com/farcasterxyz/contracts/blob/main/CONTRIBUTING.md#22-signing-commits) before they can be merged

src/NameRegistry.sol
shotaronowhere: so I hesitated to refactor variable names because I would need to update the tests, but I went ahead and implemented the variable refactoring as you suggested (updated the tests).

Honestly the view functions I added were also to make it easier to avoid changing the tests too much. I can remove those view functions and refactor the tests further. The view functions are redundant, information about the metadata of fnames can be obtained directly from the public mapping without helper view functions. The contract size is <23 kb, so the size is OK, but removing the view functions can save some contract size if other functionality is desired (24.576 kb [limit](https://ethereum.org/en/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/))

I also needed to change some function param names from `recovery` to `_recovery` to avoid a collision with the metadata name eg:

`registrationMetadataOf[tokenId].recovery = _recovery;`

This isn't an error, but solhint doesn't like it if it were written as,

`registrationMetadataOf[tokenId].recovery = recovery;`

varunsrin: 1. removing the helpers will make it easier to ship #109, so i agree about doing it 
2. thanks for updating the names! i know its a pain but it improves readability

src/NameRegistry.sol
```diff
@@ -363,10 +368,10 @@ contract NameRegistry is
     //////////////////////////////////////////////////////////////*/
 
     /**
-     * INVARIANT 1A: If an id is not minted, expiryOf[id] must be 0 and ownerOf(id) and
-     *               recoveryOf[id] must also be address(0).
+     * INVARIANT 1A: If an id is not minted, registrationMetadataOf[id].expiryTs must be 0 and ownerOf(id) and
+     *               recovery[id] must also be address(0).
```
varunsrin, line 372: this should be `registrationMetadataOf[id].recovery`

varunsrin, line 372: this should be `registrationMetadataOf[id].recovery` if i'm reading this right

src/NameRegistry.sol
```diff
@@ -383,13 +388,13 @@ contract NameRegistry is
         bytes16 fname,
         address to,
         bytes32 secret,
-        address recovery
+        address _recovery
```
varunsrin, line 391: we don't need to call this _recovery any more since you renamed the value

src/NameRegistry.sol
```diff
@@ -452,11 +457,11 @@ contract NameRegistry is
         delete timestampOf[commit];
 
         unchecked {
-            // Safety: expiryOf will not overflow given the expected sizes of block.timestamp
-            expiryOf[tokenId] = block.timestamp + REGISTRATION_PERIOD;
+            // Safety: expiry will not overflow given the expected sizes of block.timestamp
```
varunsrin, line 460: nit: expiryTs for clarity

src/NameRegistry.sol
```diff
@@ -878,17 +883,64 @@ contract NameRegistry is
...
+        if (recoveryMetadataOf[tokenId].recoveryTs == 0) revert NoRecovery();
 
         // Clear the recovery request so that it cannot be completed
-        delete recoveryClockOf[tokenId];
+        delete recoveryMetadataOf[tokenId].recoveryTs;
 
         emit CancelRecovery(sender, tokenId);
     }
 
+    /*//////////////////////////////////////////////////////////////
```
varunsrin, line 899: supportive of your suggestion of removing view functions to save space since the mappings are public and can be accessed directly

varunsrin (changes requested):
thanks pr looks great! left a few minor comments 

if you don't have time to address, i'm happy to take over and clean up.

src/NameRegistry.sol
shotaronowhere: i removed the vie funcitons and refactored the tests, saves like 300 bytes in deployed contract size

src/NameRegistry.sol
shotaronowhere: i removed the view functions and refactored the tests, saves like 300 bytes in deployed contract size

shotaronowhere:
> thanks pr looks great! left a few minor comments
> 
> if you don't have time to address, i'm happy to take over and clean up.

one last update responding to your feedback.

i'll be busy the remainder of the week, so go ahead and make any further changes you see fit to the pr

src/NameRegistry.sol
```diff
@@ -389,7 +394,7 @@ contract NameRegistry is
 
         _validateName(fname);
 
-        commit = keccak256(abi.encode(fname, to, recovery, secret));
+        commit = keccak256(abi.encodePacked(fname, to, recovery, secret));
```
varunsrin: self: why encode packed here, is this to save gas? is this safe to do without other side effects?

varunsrin (commented):
leaving some notes to self to address

test/NameRegistry.t.sol
```diff
@@ -235,6 +235,51 @@ contract NameRegistryTest is Test {
...
+        (_recoveryDestination,) = nameRegistry.recoveryMetadataOf(tokenId);
+    }
+
+    /**
+     * @notice Get the recovery clock of a fname.
+     *
+     * @param tokenId The uint256 representation of the fname
+     * @return recoveryTs The timestamp of the recovery attempt or zero if there is no active recovery.
+     */
+    function _recoveryTs(uint256 tokenId) internal view returns (uint256 recoveryTs) {
```
varunsrin, line 279: self: use consistent naming for the view functions, give them all underscores or a prefix

test/NameRegistry.t.sol
```diff
@@ -235,6 +235,51 @@ contract NameRegistryTest is Test {
         nameRegistry.makeCommit(commitHash);
     }
 
+    /*//////////////////////////////////////////////////////////////
+                            VIEW FUNCTIONS
+    //////////////////////////////////////////////////////////////*/
```
varunsrin, line 240: idea: we could make a harness contract, but probably not worth it for just four helpers

varunsrin:
@shotaronowhere can you give permissions to push to your repo

varunsrin:
Actually, I see you fixed up the encodePacked issue, so I'll merge this and clean up the rest separately. Thanks @shotaronowhere !

varunsrin (approved):

shotaronowhere:
> Actually, I see you fixed up the encodePacked issue, so I'll merge this and clean up the rest separately. Thanks @shotaronowhere !

yeah, so I added it because it is a gas optimization, but there are probably other dependencies in the repo on the commit format that I didn't update, so I removed it. Without "packed" the variables are padded with zeros to fill a 32byte slot and you pay gas for keccak256 with a base fee and per slot. By using "packed", you save a slot and save some gas without changing anything about the commit mechanic

Issue #70, opened 2022-08-27
ci: add slither


varunsrin:
adding slither to our ci would be useful for catching issues as they come up, as long as it can be executed in a reasonable amount of time and does not have too many false positives

varunsrin:
a template for this is now available here: https://github.com/ScopeLift/foundry-template

GarrettJMU:
Missed this already had a PR - happy to close mine but this should work.

https://github.com/farcasterxyz/contracts/pull/192/files

varunsrin:
Don't seem to be able to get this below 11 minutes, which may not make sense for a CI process that runs on every small change. It may be faster to just have a manual process of running slither and fixing issues before release, since its such a manual process anyway.

PR #192, opened 2023-03-03
Slither CI workflow

GarrettJMU:
## Motivation

Coming from issue:
https://github.com/farcasterxyz/contracts/issues/70

## Change Summary

Add slither to the ci workflow

## Merge Checklist

## Additional Context
n/a


varunsrin:
Thanks for making this PR. 

Unfortunately, this seems to fail because slither is taking > 10 minutes to run. I don't think its a good idea to have such an expensive test suite run on every commit to a branch. I'm going to explore some options for speeding this up, open to suggestions if you have any

GarrettJMU:
@varunsrin - Hm - I think a few other things. It doesn't look like we're not omitting node_modules but maybe taking a step back, what do you want to test here? 

Do we also need to test all the folders in lib? Seems like right now we are just omitting forge-std. I just pushed an update to omit lib altogether and might be interesting to see what time tradeoff that has.

varunsrin:
Yeah, i made another PR where I got it down to 11 minutes, but that's still painfully slow for every PR. Appreciate your work on this but it might not make sense to merge right now given how slow it is and just do the checks manually before contract releases.

varunsrin:
closing for now since we're not moving forward, thanks for trying @GarrettJMU

PR #213, opened 2023-03-06
refactor: extract TestSuite base contracts

### LINKED ISSUE #201: refactor: share helpers that are duplicated across test files
**What is the issue?**
Several helper functions to handle registrations and assertions are duplicated across test files. 

As an example, see duplication of register across [NameRegistry.t.sol ](https://github.com/farcasterxyz/contracts/blob/8ae8be1df5f19766aaf67ef5768ae0985e77e0f2/test/NameRegistry.admin.t.sol#L745)and [NameRegistry.admin.t.sol](https://github.com/farcasterxyz/contracts/blob/8ae8be1df5f19766aaf67ef5768ae0985e77e0f2/test/NameRegistry.t.sol#L2476)

horsefacts:
## Motivation

Extract shared `TestSuite` base contracts for existing tests. Close #201.

## Change Summary

Extract abstract base `TestSuite` contracts to share setup and examples across similar tests. Reorganize test files to follow this convention.

## Merge Checklist

## Additional Context

This is a middle-of-the-road option for sharing duplicated helpers: each contract has its own base `TestSuite` contract containing helpers and setup. 

Alternatives:
- Share even more with a single base `TestSuite` contract that may contain unused setup, constants, and helpers for some test contracts.
- Share less by extracting helpers as free functions. Since most of these make use of instance variables they would need to be passed as args.

Note: changing tests changed the gas snapshot baselines, so this PR includes a new `.gas-snapshot`.


.gas-snapshot
```diff
@@ -1,6 +1,201 @@
 BundleRegistryGasUsageTest:testGasRegister() (gas: 1630351)
 BundleRegistryGasUsageTest:testGasTrustedRegister() (gas: 1468291)
+BundleRegistryTest:testFuzzCannotChangeTrustedCallerToZeroAddress(address) (runs: 512, μ: 17362, ~: 17362)
```
varunsrin, line 3: I run snapshot on a small selection of tests that model typical gas consumption and ignore the rest of the tests since they tend to change everytime you change the test logic (which is often) 

Running `forge snapshot --match-contract Gas` should resolve this issue

varunsrin (changes requested):
I like the refactor, cleans things up nicely. 

Only 1 change re: the gas snapshot and then we're good to merge if tests are passing.

.gas-snapshot
horsefacts: Aha, that makes sense. Fixed this up.

varunsrin (approved):

Issue #209, opened 2023-03-05
test(NameRegistry): test that the roles in the contract persist correctly after an upgrade


varunsrin:
**What is the test?**
In the NameRegistryUpgrade test, assign roles and ensure that they persist correctly even after the contract has been upgraded

varunsrin:
Closing since NameRegistry was removed: https://github.com/farcasterxyz/contracts/pull/224

PR #214, opened 2023-03-07
test: NameRegistry upgrade should maintain existing roles

bstchow:
## Motivation

This closes https://github.com/farcasterxyz/contracts/issues/209

## Change Summary

Adds a test that NameRegistry roles are maintained after an upgrade to a new implementation contract.

## Merge Checklist

### DIFF
--- a/test/NameRegistry/NameRegistry.upgrade.t.sol
+++ b/test/NameRegistry/NameRegistry.upgrade.t.sol
@@ -59,6 +59,22 @@ contract NameRegistryUpgradeTest is NameRegistryTestSuite {
         assertEq(nameRegistryV2.number(), 42);
     }
 
+    function testUpgradeMaintainsExistingRoles(bytes32 role1, address alice, address bob) public {
+        vm.assume(bob != alice);
+        _grant(role1, alice);
+
+        assertTrue(nameRegistry.hasRole(role1, alice), "alice has role1");
+        assertFalse(nameRegistry.hasRole(role1, bob), "bob does not have role1");
+
+        NameRegistryV2 nameRegistryV2 = _upgradeToV2();
+
+        vm.prank(ADMIN);
+        nameRegistryV2.initializeV2("Farcaster NameRegistry", "FCN", VAULT, POOL);
+
+        assertTrue(nameRegistryV2.hasRole(role1, alice), "alice still has role1");
+        assertFalse(nameRegistryV2.hasRole(role1, bob), "bob still does not have role1");
+    }
+
     // solhint-disable-next-line no-empty-blocks
     function testModifiedFnAfterUpgrade() public {
         // TODO: If we decide to stick with UUPS, assert that a function can be redefined and will
@@ -99,7 +115,7 @@ contract NameRegistryV2 is
     Initializable,
     ERC721Upgradeable,
     ERC2771ContextUpgradeable,
-    OwnableUpgradeable,
+    AccessControlUpgradeable,
     PausableUpgradeable,
     UUPSUpgradeable
 {
@@ -144,9 +160,6 @@ contract NameRegistryV2 is
 
         __ERC721_init_unchained(_name, _symbol);
 
-        // Initialize the owner to the deployer and then transfer it to _owner
-        __Ownable_init_unchained();
-
         vault = _vault;
 
         pool = _pool;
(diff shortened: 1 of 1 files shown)


bstchow:
Seems like label editing privs are off limits for first time contributors

test/NameRegistry.upgrade.t.sol
```diff
@@ -108,6 +139,7 @@ contract NameRegistryV2 is
     Initializable,
     ERC721Upgradeable,
     ERC2771ContextUpgradeable,
+    AccessControlUpgradeable,
```
varunsrin, line 142: good catch, though i think OwnableUpgradeable in the next line can be removed

test/NameRegistry.upgrade.t.sol
```diff
@@ -68,6 +68,31 @@ contract NameRegistryUpgradeTest is Test {
...
+        bytes32 role2,
+        bytes32 role3,
+        address alice,
+        address bob,
+        address charlie
+    ) public {
+        _grant(role1, alice);
+        _grant(role2, bob);
+
+        vm.assume(!nameRegistryProxy.hasRole(role3, charlie));
```
varunsrin, line 82: nit: i think you can simplify it to just a single role, and alice having the role and bob not having one, unless there is some codepath that won't be tested if we simplify

test/NameRegistry.upgrade.t.sol
```diff
@@ -190,4 +222,13 @@ contract NameRegistryV2 is
     function setNumber(uint256 _number) public {
         number = _number;
     }
+
+    function supportsInterface(bytes4 interfaceId)
```
varunsrin, line 226: does this break a test if not included?

varunsrin (changes requested):
Overall looks great, left some feedback

test/NameRegistry.upgrade.t.sol
bstchow: I think the same codepaths would be tested if we omitted a second granted role, but I think testing the successful migration of multiple storage slots is a bit more implementation agnostic, although I agree paranoid 😂

test/NameRegistry.upgrade.t.sol
bstchow: I believe both AccessControlUpgradeable & ERC721Upgradeable require supportsInterface so I think it needs to be defined here as well

```
error[6480]: TypeError: Derived contract must override function "supportsInterface". Two or more base classes define function with same name and parameter types.
   --> test/NameRegistry.upgrade.t.sol:138:1:
```

test/NameRegistry.upgrade.t.sol
bstchow: Done 👍

test/NameRegistry.upgrade.t.sol
```diff
@@ -183,11 +201,20 @@ contract NameRegistryV2 is
         return ERC2771ContextUpgradeable._msgData();
     }
 
-    // solhint-disable-next-line no-empty-blocks
-    function _authorizeUpgrade(address) internal override onlyOwner {}
+    // // solhint-disable-next-line no-empty-blocks
```
varunsrin, line 204: nit: double comments

test/NameRegistry.upgrade.t.sol
varunsrin: 👍

test/NameRegistry.upgrade.t.sol
```diff
@@ -99,6 +114,12 @@ contract NameRegistryUpgradeTest is Test {
         nameRegistryProxy.upgradeTo(address(nameRegistryV2Impl));
         return NameRegistryV2(address(proxy));
     }
+
+    /// TODO: Should be moved to shared test suite for NameRegistry. (GH Issue #209)
```
varunsrin, line 118: I think this can be moved into NameRegistryTestSuite and the todo can be resolved

test/NameRegistry.upgrade.t.sol
```diff
@@ -68,6 +68,21 @@ contract NameRegistryUpgradeTest is Test {
         assertEq(nameRegistryV2Proxy.number(), 42);
     }
 
+    function testUpgradeMaintainsExistingRoles(bytes32 role1, bytes32 role2, address alice, address bob) public {
+        _grant(role1, alice);
+
+        vm.assume(!nameRegistryProxy.hasRole(role2, bob));
```
varunsrin, line 74: why assume here instead of just asserting?

varunsrin (changes requested):
left a few minor comments, and the pr needs a rebase

test/NameRegistry.upgrade.t.sol
bstchow: True - will assert

varunsrin:
Closing because we've [deprecated the NameRegistry](https://github.com/farcasterxyz/contracts/pull/224)

PR #215, opened 2023-03-07
chore: remove yarn, node, husky; add rusty-hook

### LINKED ISSUE #199: feat: use a cargo/rust based system for hooks and deprecate npm/husky
**What is the feature you would like to implement?**

Instead of using husky to manage our git hooks for running tests before commits, use a cargo based system to manage this. 

**Why is this feature important?**

Allows us to eliminate yarn and javascript from our codebase.  

**Will the protocol spec need to be updated??**
No

appleseed-iii:
## Motivation

Fixes #199 - uses cargo/rust for pre-commit hooks rather than yarn, node, husky.

## Change Summary

added [rusty-hook](https://github.com/swellaby/rusty-hook), removed yarn.lock & package.json. rusty-hook adds shell scripts in the pre-existing `.husky` dir. Updated README to add Rust/Cargo as a dev dependency.

## Merge Checklist


.gitignore
```diff
@@ -18,4 +18,19 @@ yarn-debug.log*
...
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
+# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
+Cargo.lock
+
+# These are backup files generated by rustfmt
+**/*.rs.bk
+
+# MSVC Windows builds of rustc generate these, which store debugging information
+*.pdb
```
appleseed-iii, line 36: added rust related items to gitignore, but left the original yarn items for backwards compatibility.

.husky/pre-commit
```diff
@@ -1,4 +1,25 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-forge snapshot --check --match-contract Gas
-npx lint-staged
```
appleseed-iii, line 4: rusty-hook overwrites the existing shell scripts from husky.

.rusty-hook.toml
```diff
@@ -0,0 +1,5 @@
+[hooks]
+pre-commit = "forge fmt && forge test -vvv && forge snapshot --check --match-contract Gas"
```
appleseed-iii, line 2: running `forge fmt` locally so that linting is fixed pre-commit/before the CI.

Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "2.0.0"
+edition = "2021"
+publish = false
+
+[[bin]]
+name = "main"
+path = "main.rs"
+
+[dev-dependencies]
+rusty-hook = "^0.11.2"
```
appleseed-iii, line 12: should be the same necessities from obsolete `package.json`

main.rs
```diff
@@ -0,0 +1,3 @@
+// an empty cargo target
+// just allows rusty-hook to run
+fn main() {}
```
appleseed-iii, line 3: `cargo build` (which installs rusty-hook locally) cannot run without a build target, so I made an empty main func

appleseed-iii (commented):
summarizing the nuances

.rusty-hook.toml
appleseed-iii, line 2: i need to revise the last shell command `forge snapshot --check --match-contract Gas`; it won't allow commits with a LOWER gas snapshot

varunsrin:
thanks, appreciate the thorough summary.

varunsrin (commented):
Two questions: 

1. It seems odd that shell scripts for rusty hook are in the `.husky` folder, which will be confusing to new readers.

2. Do we need to have all the shell scripts even though we are not using them? husky would only selectively install them as needed.

appleseed-iii:
> Two questions: 
> 
> 
> 
> 1. It seems odd that shell scripts for rusty hook are in the `.husky` folder, which will be confusing to new readers.
> 
> 
> 
> 2. Do we need to have all the shell scripts even though we are not using them? husky would only selectively install them as needed. 

1. I thought about this, too. Ultimately I figured most devs prob don't even look in the `.husky` directory bc technically it's a hidden directory and it's auto generated / never edited by the dev (just like `.git`). I'll double check if that's a config setting I can easily change.

2. I can def change the config for this and will fix

build.rs
```diff
@@ -0,0 +1,11 @@
...
+
+// remove husky configuration from .git/config if it exists
+fn main() {
+    Command::new("git")
+        .arg("config")
+        .arg("--unset")
+        .arg("core.hooksPath")
+        .status()
+        .expect("core.hooksPath failed to reset. You should manually run git config --unset core.hooksPath");
+}
```
appleseed-iii, line 11: So with this update when a dev runs `cargo build` (per the directions in CONTRIBUTING.md) the following steps occur in this order:
1. `build.rs` runs & removes `hooksPath = .husky" from the dev's git config (per the above)
2. rusty-hook is installed as a dependency per `Cargo.toml`
3. rusty-hook's post-build script installs the git commit hooks in the hidden `.git/hooks` dir

So, technically, I didn't reduce the number of commit hooks that are installed, but they are hidden in the `.git/hooks` folder as they should be.

I *could* remove the extra git hooks, but I would need to add more commands to `.build.rs` & it feels like overkill.

varunsrin: oh nice, this is great. i like this solution

PR #217, opened 2023-03-27
test: `IdRegistry` invariants

horsefacts:
🚧 

- Add invariant test handler/tests for `IdRegistry`
- Add `LibAddressSet` helper
- Add invariant test configuration
- Update `forge-std`

### DIFF
--- a/test/IdRegistry/IdRegistry.invariants.t.sol
+++ b/test/IdRegistry/IdRegistry.invariants.t.sol
@@ -57,4 +57,19 @@ contract IdRegistryInvariants is IdRegistryTestSuite {
             }
         }
     }
+
+    mapping(uint256 => bool) seenFids;
+
+    /// @notice A nonzero address may only own one fid at a time. All fids must have a unique owner.
+    function invariant_4_allFidsHaveUniqueOwner() public {
+        address[] memory fidOwners = handler.fidOwners();
+        for (uint256 i; i < fidOwners.length; ++i) {
+            address owner = fidOwners[i];
+            uint256 fid = idRegistry.idOf(owner);
+            if (seenFids[fid]) {
+                revert("invariant 4");
+            }
+            seenFids[fid] = true;
+        }
+    }
 }
--- a/test/IdRegistry/IdRegistry.invariants.t.sol
+++ b/test/IdRegistry/IdRegistry.invariants.t.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.18;
+
+import {IdRegistryTestSuite} from "./IdRegistryTestSuite.sol";
+import {IdRegistryHandler} from "./handlers/IdRegistryHandler.sol";
+
+/* solhint-disable state-visibility */
+
+contract IdRegistryInvariants is IdRegistryTestSuite {
+    IdRegistryHandler handler;
+
+    function setUp() public override {
+        super.setUp();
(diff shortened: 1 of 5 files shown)


test/IdRegistry/IdRegistry.invariants.t.sol
```diff
@@ -0,0 +1,60 @@
...
+        super.setUp();
+        idRegistry.disableTrustedOnly();
+        handler = new IdRegistryHandler(idRegistry, address(this));
+
+        targetContract(address(handler));
+    }
+
+    /// @notice All recovery addresses must be associated with a fid.
+    function invariant_1_allRecoveryAddrsAssociatedWithFid() public {
+        address[] memory recoveryAddrs = handler.recoveryAddrs();
```
varunsrin, line 22: The tricky part about using the handler here is that it is duplicating the logic of the contract to allow testability, but this might not protect against someone introducing a bug in the contract logic later, or if there are bugs in reimplementing the contract logic in the handler. 

Consider the case where [this line](https://github.com/farcasterxyz/contracts/blob/78c205a7e0ab26de151e43c7ccf0198f968ddd29/src/IdRegistry.sol#L279) was accidentally removed, the invariant tests would still pass.

varunsrin, line 22: The tricky part about using the handler here is that it is duplicating the logic of the contract to allow testability, but this might not protect against someone introducing a bug in the contract logic later, or if there are bugs in reimplementing the contract logic in the handler. 

Consider the case where [this line](https://github.com/farcasterxyz/contracts/blob/78c205a7e0ab26de151e43c7ccf0198f968ddd29/src/IdRegistry.sol#L279) was accidentally removed, the invariant tests would still pass.

varunsrin (commented):
Left a few comments on this, thanks for getting the framework set up

test/IdRegistry/IdRegistry.invariants.t.sol
horsefacts: Yeah, I agree. The amount of accounting we have to do in the handler is not ideal and a bit of a code smell. I'm going to try replacing some of the state in the handler with stateless calculations. For example, since we know the latest fid, we can iterate over all fids and build up `_fidsByRecoveryAddr` without needing to duplicate the accounting. I think we can do the same for the reverse fid->owner mapping: rather than keep track of it in the handler, we can iterate over all "known addresses" and construct it on demand.

test/IdRegistry/IdRegistry.invariants.t.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.18;
+
+import {IdRegistryTestSuite} from "./IdRegistryTestSuite.sol";
+import {IdRegistryHandler} from "./handlers/IdRegistryHandler.sol";
+
+/* solhint-disable state-visibility */
+
+contract IdRegistryInvariants is IdRegistryTestSuite {
```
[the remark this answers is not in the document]
horsefacts: Good one. This is equivalent to "every registered fid must have a unique owner address," right? I think that's easier to test.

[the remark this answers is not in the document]
varunsrin: yeah that statement seems equivalent for our purposes

varunsrin:
Closing since our invariants have changed

PR #219, opened 2023-06-01
fix: reorder expectEmit

horsefacts:
## Motivation

Foundry 1.0 made a [breaking change](https://github.com/foundry-rs/foundry/blob/master/CHANGELOG.md#breaking-changes) to the behavior of `vm.expectEmit` so that it only applies to the next call. This means tests are now more sensitive to the order in which the expectation, `emit`, and call are defined. Looks like there's one test in the FC contracts that's affected.

## Change Summary

Reorder call in `IdRegistryOwnertest#FuzzDisableTrustedCaller`.

## Merge Checklist

## Additional Context

An alternative fix is to pin the latest stable pre-1.0 Foundry in CI following the instructions [here](https://github.com/foundry-rs/foundry/blob/master/CHANGELOG.md#breaking-changes) rather than keep up with Foundry nightly.

### DIFF
--- a/test/IdRegistry/IdRegistry.owner.t.sol
+++ b/test/IdRegistry/IdRegistry.owner.t.sol
@@ -51,8 +51,8 @@ contract IdRegistryOwnerTest is IdRegistryTestSuite {
         assertEq(idRegistry.getTrustedOnly(), 1);
 
         vm.expectEmit(true, true, true, true);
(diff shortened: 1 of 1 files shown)


varunsrin:
Sorry, missed this PR and ended up fixing it here: https://github.com/farcasterxyz/contracts/pull/221

PR #220, opened 2023-06-16
feat: add StorageRent contract

horsefacts:
## Motivation

Initial version of an onchain storage rental contract. [Related FIP](https://github.com/farcasterxyz/protocol/discussions/98).
## Change Summary

- Add `StorageRent` contract and tests.
- Install Chainlink price/uptime feed dependencies.
- Add `MockPriceFeed` and `MockUptimeFeed` in `test/Utils.sol`. Use these to stub Chainlink feed data.
- Add a `RevertOnReceive` helper to `test/Utils.sol`. Use this to test reverting ether transfers.

## Merge Checklist

## Additional Context
- Adds a new dependency, [`smartcontractkit/chainlink-brownie-contracts`](https://github.com/smartcontractkit/chainlink-brownie-contracts). Although "Brownie" is in the name, this is the official Chainlink repo for tools that install from Github, including Foundry.
- Docs on Chainlink's L2 [sequencer uptime feeds](https://docs.chain.link/data-feeds/l2-sequencer-feeds).
- This PR includes the test fix from https://github.com/farcasterxyz/contracts/pull/219.

---

## PR-Codex overview
This PR adds a new `StorageRent` contract and its associated tests, which allows users to rent storage space on the Ethereum network for a fee.

### Detailed summary
- Added `StorageRent` contract with functionality for renting storage space
- Added tests for `StorageRent` contract and associated utility contracts
- Added `MockChainlinkFeed`, `MockPriceFeed`, and `MockUptimeFeed` contracts for testing purposes
- Added `FixedPointMathLib` library for fixed point math calculations
- Added several error messages and events for better user feedback and contract functionality

> The following files were skipped due to too many changes: `src/StorageRent.sol`, `test/StorageRent/StorageRent.t.sol`


src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+                                 ERRORS
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev Revert if the caller attempts to rent storage after the rental period has ended.
+    error RentalPeriodHasEnded();
+
+    /// @dev Revert if the caller attempts to rent more storage than is available.
+    error ExceedsCapacity();
+
+    /// @dev Revert if the caller attempts a batch rent with mismatched input array lengths.
```
horsefacts, line 18:
```suggestion
    /// @dev Revert if the caller attempts a batch rent with mismatched input array lengths or an empty array.
```

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+    /// @dev Revert if the price feed round is incomplete and has not yet generated a price.
+    error IncompleteRound();
+
+    /// @dev Revert if the price feed returns a zero or negative price.
+    error InvalidPrice();
+
+    /// @dev Revert if the sequencer uptime feed detects that the L2 sequencer is unavailable.
+    error SequencerDown();
+
+    /// @dev Revert if the L2 sequencer restarted less than SEQUENCER_RESTART_GRACE_PERIOD seconds ago.
```
horsefacts, line 42:
```suggestion
    /// @dev Revert if the L2 sequencer restarted less than L2_DOWNTIME_GRACE_PERIOD seconds ago.
```

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+     * @dev Chainlink L2 sequencer uptime feed.
+     */
+    AggregatorV3Interface public immutable uptimeFeed;
+
+    /*//////////////////////////////////////////////////////////////
+                                 STORAGE
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @dev Time at which this contract will no longer accept storage rent payments. Changeable by owner.
```
horsefacts, line 120:
```suggestion
     * @dev Block timestamp at which this contract will no longer accept storage rent payments. Changeable by owner.
```

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+    /*//////////////////////////////////////////////////////////////
+                               CONSTRUCTOR
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Set the price feed, uptime feed, and initial parameters.
+     *
+     * @param _priceFeed  Chainlink ETH/USD price feed.
+     * @param _uptimeFeed Chainlink L2 sequencer uptime feed.
+     */
```
horsefacts, line 148: Missing additional params here.

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+        if (block.timestamp >= rentalPeriodEnd) revert RentalPeriodHasEnded();
+        if (fids.length == 0 || units.length == 0) revert InvalidBatchInput();
+        if (fids.length != units.length) revert InvalidBatchInput();
+
+        uint256 totalCost;
+        for (uint256 i; i < fids.length; ++i) {
+            uint256 qty = units[i];
+            if (qty == 0) continue;
+            if (rentedUnits + qty > maxUnits) revert ExceedsCapacity();
+            totalCost += price(qty);
```
horsefacts, line 199: Gas: refactor to cache price feed data outside the loop.

test/Utils.sol
```diff
@@ -83,3 +86,85 @@ contract BundleRegistryHarness is BundleRegistry {
...
+        return latestRoundData();
+    }
+
+    function latestRoundData() public view returns (uint80, int256, uint256, uint256, uint80) {
+        return
+            (roundData.roundId, roundData.answer, roundData.startedAt, roundData.timeStamp, roundData.answeredInRound);
+    }
+}
+
+contract MockUptimeFeed is AggregatorV3Interface {
```
horsefacts, line 133: These are really similar, extract a base contract.

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+    /**
+     * @notice Cost in wei to rent one storage unit.
+     */
+    function unitPrice() public view returns (uint256) {
+        return price(1);
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                              OWNER ACTIONS
+    //////////////////////////////////////////////////////////////*/
```
horsefacts, line 243: TODO: Do we need admin function for free storage for existing fids?

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+                                 EVENTS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @dev Emit an event when
+     *
+     * @param payer     Address of the account paying the storage rent.
+     * @param fid       The fid that will receive the storage allocation.
+     * @param timestamp block.timestamp of the rent transaction
+     *                  (TODO: Do we need this in the event or can we just get it from the block?)
```
horsefacts, line 55: TODO here: can probably omit timestamp here.

src/StorageRegistry.sol
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                              PRICE VIEWS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Calculate the cost in wei to rent the given number of storage units.
+     *
+     * @param units Number of storage units.
```
horsefacts, line 215: TODO: missing return parameter natspec.

horsefacts: (all over 😅)

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+    event Withdraw(address indexed to, uint256 amount);
+
+    /*//////////////////////////////////////////////////////////////
+                                CONSTANTS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @dev Period in seconds to wait after the L2 sequencer restarts before resuming rentals.
+     */
+    uint256 public constant L2_DOWNTIME_GRACE_PERIOD = 3600;
```
varunsrin, line 97: Would be helpful to have some comments here about why this number is the right tradeoff for us

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.18;
```
varunsrin, line 2: A few nits on contract versioning: 

1. Should we use standard versioning (v1, v2) or stick to the date versioning (2023.5.31) that we use for the protocol? Initial reaction is that having the date versions makes it easier to know which contract maps to which release, but I hold this weakly. 

2. Should the version number go into the name of the contract? StorageRegistryv1 instead of StorageRegistry

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        usdUnitPrice = usdPrice;
+    }
+
+    /**
+     * @notice Change the maximum supply of storage units.
+     *
+     * @param max The new maximum supply of storage units.
+     */
+    function setMaxUnits(uint256 max) external onlyOwner {
+        emit SetMaxUnits(maxUnits, max);
```
varunsrin, line 295: do we want a hardcoded upper bound to prevent overflow?

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Rent storage for a given fid. The caller must provide an exact payment amount.
+     *
+     * @param fid   The fid that will receive the storage allocation.
+     * @param units Number of storage units to rent.
+     */
+    function rent(uint256 fid, uint256 units) external payable whenNotDeprecated {
+        if (msg.value != price(units)) revert InvalidPayment();
```
varunsrin, line 184: Should we accept overpayments and return the excess funds? The concern is that we might be reverting transactions and frustrating users who used a slightly higher exchange rate

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        for (uint256 i; i < fids.length; ++i) {
+            uint256 qty = units[i];
+            if (qty == 0) continue;
+            if (rentedUnits + qty > maxUnits) revert ExceedsCapacity();
+            totalCost += _price(qty, _usdUnitPrice, _ethUsdPrice);
+            rentedUnits += qty;
+            emit Rent(msg.sender, fids[i], qty);
+        }
+
+        if (msg.value != totalCost) revert InvalidPayment();
```
varunsrin, line 215: same comment about accepting overpayments

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        return _price(units, usdUnitPrice, ethUsdPrice());
+    }
+
+    function ethUsdPrice() public view returns (uint256) {
+        /* Ensure that the L2 sequencer is up. */
+        (, int256 sequencerUp, uint256 startedAt,,) = uptimeFeed.latestRoundData();
+        if (sequencerUp != 0) revert SequencerDown();
+
+        /* If the L2 sequencer recently restarted, ensure the grace period has elapsed. */
+        uint256 timeSinceUp = block.timestamp - startedAt;
```
varunsrin, line 244: Is it possible that `startedAt` from the uptime feed could return zero or some incorrect value?

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        (uint80 roundId, int256 answer,, uint256 updatedAt, uint80 answeredInRound) = priceFeed.latestRoundData();
+        if (answer <= 0) revert InvalidPrice();
+        if (updatedAt == 0) revert IncompleteRound();
+        if (answeredInRound < roundId) revert StalePrice();
+
+        return uint256(answer);
+    }
+
+    function _price(uint256 units, uint256 usdPerUnit, uint256 ethPerUsd) internal pure returns (uint256) {
+        return units * usdPerUnit * 1e18 / ethPerUsd;
```
varunsrin, line 257: should we be protecting against overflows here if the `units` value is set to something really large in the batch method?

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        deprecationTimestamp = timestamp;
+    }
+
+    /**
+     * @notice Withdraw a specified amount of ether from the contract balance to a given address.
+     *
+     * @param to     Address of recipient.
+     * @param amount The amount of ether to withdraw.
+     */
+    function withdraw(address to, uint256 amount) external onlyOwner {
```
varunsrin, line 315: security: thoughts on having two levels of admins? one that can move funds (for operations) and one that can change where the funds go (vault pattern)

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        maxUnits = max;
+    }
+
+    /**
+     * @notice Change the deprecationTimestamp.
+     *
+     * @param timestamp The new deprecationTimestamp.
+     */
+    function setDeprecationTimestamp(uint256 timestamp) external onlyOwner {
+        emit SetDeprecationTimestamp(deprecationTimestamp, timestamp);
```
varunsrin, line 305: should we prevent this from being set in the past?

src/StorageRegistry.sol
```diff
@@ -0,0 +1,321 @@
...
+        /* Ensure that the L2 sequencer is up. */
+        (, int256 sequencerUp, uint256 startedAt,,) = uptimeFeed.latestRoundData();
+        if (sequencerUp != 0) revert SequencerDown();
+
+        /* If the L2 sequencer recently restarted, ensure the grace period has elapsed. */
+        uint256 timeSinceUp = block.timestamp - startedAt;
+        if (timeSinceUp < L2_DOWNTIME_GRACE_PERIOD) revert GracePeriodNotOver();
+
+        /* Get and validate the Chainlink ETH/USD price. */
+        (uint80 roundId, int256 answer,, uint256 updatedAt, uint80 answeredInRound) = priceFeed.latestRoundData();
```
varunsrin, line 248: Is it possible that `answer` could be set super low or high due to an error upstream? Wondering if we should guard with some sane boundaries.

src/StorageRegistry.sol
horsefacts: If the contracts are coupled to a protocol version, I think the same format makes sense for both. I'd prefer adding a `VERSION` constant to each contract rather than using the contract name (especially if we're using the date-based protocol version format).

src/StorageRegistry.sol
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * @dev Total capacity of storage units. Changeable by owner.
+     */
+    uint256 public maxUnits;
+
+    /**
+     * @dev Total number of storage units that have been rented.
+     */
+    uint256 public rentedUnits;
```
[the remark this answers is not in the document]
horsefacts: (revisit once all the params are stable)

src/StorageRegistry.sol
```diff
@@ -0,0 +1,422 @@
...
+                              OWNER ACTIONS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Credit multiple fids with free storage units. Only callable by owner.
+     *
+     * @param fids  An array of fids.
+     * @param units Number of storage units per fid.
+     */
+    function batchCredit(uint256[] calldata fids, uint256 units) external onlyOwner whenNotDeprecated {
```
horsefacts, line 353: - Probably want a single-fid version for testnet
- MetaTX support?

src/StorageRegistry.sol
horsefacts: Decided to parameterize this rather than use a constant because I don't have a great answer. Seems like something we might want the ability to change. This is the value used in the Chainlink docs—I'll read up on this further.

src/StorageRegistry.sol
horsefacts: Checked math should keep this multiplication and the accumulator in `batchRent` safe against overflow, but we might revert early with a checked math error during the price calculation in `batchRent` if the caller passes a very high `units` value. The error is not super friendly, but it's safe. This seems OK to me, what do you think?

src/StorageRegistry.sol
horsefacts: Definitely possible, historically unlikely, there's a first time for everything. I think a good design is to trigger a circuit breaker if the price hits some min/max bound. 

Zooming out, I think the bigger question is what we want the failure mode to be for hubs, clients, and users if storage rentals are down because a dependency (sequencer, price feed) is broken.

varunsrin (commented):
Reviewed the `rent` function and tests and left feedback. Will review the rest shortly.

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1134 @@
...
+        newEthUsdPrice = bound(newEthUsdPrice, 1, type(int256).max);
+
+        rentStorage(msgSender1, id1, units1);
+
+        // Set a new ETH/USD price
+        priceFeed.setPrice(newEthUsdPrice);
+
+        vm.warp(block.timestamp + fcStorage.priceFeedCacheDuration() + 1);
+
+        rentStorage(msgSender2, id2, units2);
```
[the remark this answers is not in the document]
horsefacts: yeah, good call.

src/StorageRegistry.sol
varunsrin: that seems reasonable, i forget math is default checked now

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1134 @@
...
+        int256 newEthUsdPrice,
+        uint256 warp
+    ) public {
+        uint256 lastPriceFeedUpdate = fcStorage.lastPriceFeedUpdate();
+        uint256 ethUsdPrice = fcStorage.ethUsdPrice();
+
+        // Ensure Chainlink price is positive
+        newEthUsdPrice = bound(newEthUsdPrice, 1, type(int256).max);
+
+        rentStorage(msgSender1, id1, units1);
```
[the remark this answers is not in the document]
horsefacts: it is possible, but the second call to `rentStorage` will throw away the run at `vm.assume(remaining > 0)`.

src/StorageRent.sol
```diff
@@ -0,0 +1,578 @@
...
+    function withdraw(uint256 amount) external onlyTreasurer {
+        emit Withdraw(vault, amount);
+        _sendNative(vault, amount);
+    }
+
+    /**
+     * @notice Native token transfer helper.
+     */
+    function _sendNative(address to, uint256 amount) internal {
+        if (address(this).balance < amount) revert InsufficientFunds();
```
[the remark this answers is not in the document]
horsefacts: Just clarity and consistency—this was the same transfer code that was in the `NameRegistry`, and itself is close to [what OZ does](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e50c24f5839db17f46991478384bfda14acfb830/contracts/utils/Address.sol#L65).

If we omit this, it will just revert with `CallFailed` instead.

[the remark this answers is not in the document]
horsefacts: I guess this does have a gas impact in some cases too, since it avoids trying to make the external call.

src/StorageRent.sol
```diff
@@ -0,0 +1,578 @@
...
+        if (rentedUnits + units > maxUnits) revert ExceedsCapacity();
+        uint256 totalPrice = _price(units);
+        if (msg.value < totalPrice) revert InvalidPayment();
+
+        // Effects
+        rentedUnits += units;
+        emit Rent(msg.sender, fid, units);
+
+        // Interactions
+        if (msg.value > totalPrice) {
```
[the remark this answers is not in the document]
horsefacts: Hmm, I think I'd expect most callers to retrieve the latest `price(units)` and then `rent(fid, units)` with the exact amount returned, rather than add some safety buffer. Allowing overpayments means we gracefully handle the case when the ETH/USD price happened to decrease intra-block or while the tx was waiting. But if the price goes *up* we'll still revert.  

That should be pretty infrequent, at least for users who aren't lowballing gas: to revert, the tx has to be in the same 2-second block as an earlier price feed update that increases the price.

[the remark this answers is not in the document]
horsefacts: Actually, we could probably handle this case too, if we use block number to make the "honor old price" decision: hold on to the previous price, store the block number of the last change, and use the old price if it changed in this block. Then, the price change would only apply starting in the next block, rather than the next tx.

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+    function testFuzzUnitPriceCached(uint48 usdUnitPrice, int256 ethUsdPrice) public {
+        // Ensure Chainlink price is positive
+        ethUsdPrice = bound(ethUsdPrice, 1, type(int256).max);
+
+        uint256 cachedPrice = fcStorage.ethUsdPrice();
+
+        priceFeed.setPrice(ethUsdPrice);
+
+        vm.prank(admin);
+        fcStorage.setPrice(usdUnitPrice);
```
varunsrin, line 720: should we call refresh here to ensure that calling it when the cached price is valid doesn't trigger any unwanted side effects?

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+
+        assertEq(fcStorage.unitPrice(), uint256(usdUnitPrice) * 1e18 / cachedPrice);
+    }
+
+    function testPriceRoundsUp() public {
```
varunsrin, line 725: nit: i didn't quite understand what this test was doing

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+                answeredInRound: 1
+            })
+        );
+
+        vm.expectRevert(StorageRent.GracePeriodNotOver.selector);
+        vm.prank(admin);
+        fcStorage.refreshPrice();
+    }
+
+    function testFuzzOnlyAuthorizedCanRefreshPrice(address caller) public {
```
varunsrin, line 876: do we ever explicitly test that the admin can call this function? i know we do it implicitly in a bunch of places, but wondering if we should have an explicit check like we do with FuzzCredit

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+        }
+        vm.prank(msgSender);
+        fcStorage.batchRent{value: totalCost}(ids, units);
+
+        // Expect rented units to increase
+        assertEq(fcStorage.rentedUnits(), rented + totalUnits);
+        return totalCost;
+    }
+
+    function credit(address msgSender, uint256 id, uint256 units) public {
```
varunsrin, line 1263: for the test methods `credit` and `rentStorage` I was a little confused about what they did when I read them in code. 

their purpose is to assert that a certain function works with side effects, so maybe we should call them `assertCredit` and `assertRentStorage` so that their purpose is more clear?

script/StorageRent.s.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.18;
```
varunsrin, line 2: thoughts on upgrading all our contracts to 8.20?

src/StorageRent.sol
```diff
@@ -0,0 +1,625 @@
...
+            _sendNative(msg.sender, msg.value - totalPrice);
+        }
+    }
+
+    /**
+     * @notice Rent storage for multiple fids. The caller must provide an exact payment amount equal to
```
varunsrin, line 343: exact payment is no longer necessary

src/StorageRent.sol
```diff
@@ -0,0 +1,625 @@
...
+     *         the sum of the prices for each fid's storage allocation.
+     *
+     * @param fids  An array of fids.
+     * @param units An array of storage unit quantities. Must be the same length as the fids array.
+     */
+    function batchRent(uint256[] calldata fids, uint256[] calldata units) external payable whenNotDeprecated {
+        // Pre-checks
+        if (fids.length == 0 || units.length == 0) revert InvalidBatchInput();
+        if (fids.length != units.length) revert InvalidBatchInput();
```
varunsrin, line 352: minor gas optimization, we avoid one .length lookup with: 

```
        // Pre-checks
        if (fids.length != units.length) revert InvalidBatchInput();
        if (fids.length == 0) revert InvalidBatchInput();
```

script/StorageRent.s.sol
horsefacts: Depends if our target L2 supports `PUSH0`. I don't think Optimism/Arbitrum do yet, but I'll double check.

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+        assertEq(unitPrice1, unitPrice2);
+        assertEq(unitPricePaid1, unitPricePaid2);
+        assertEq(unitPricePaid1, unitPrice1);
+        assertEq(fcStorage.lastPriceFeedUpdateTime(), block.timestamp);
+        assertEq(fcStorage.lastPriceFeedUpdateBlock(), block.number);
+        assertEq(fcStorage.prevEthUsdPrice(), ethUsdPrice);
+        assertEq(fcStorage.ethUsdPrice(), newEthUsdPrice);
+    }
+
+    function testFuzzRentRevertsAfterDeadline(address msgSender, uint256 id, uint256 units) public {
```
varunsrin, line 234: what about WhenDeprecated instead of AfterDeadline ?

src/StorageRent.sol
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    /**
+     * @notice Credit a continuous sequence of fids with free storage units. Only callable by operator.
+     *
+     * @param start Lowest fid in sequence (inclusive).
+     * @param end   Highest fid in sequence (inclusive).
+     * @param units Number of storage units per fid.
+     */
+    function continuousCredit(uint256 start, uint256 end, uint256 units) external onlyOperator whenNotDeprecated {
```
varunsrin, line 523: if we won't ever realistically use this after the initial seed, should we just always have start == 1? that simplifies this logic even further

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+        uint256 units = bound(_units, 1, type(uint32).max);
+        uint256 end = uint256(start) + bound(n, 1, 10000);
+        vm.assume(caller != operator);
+
+        vm.prank(caller);
+        vm.expectRevert(StorageRent.NotOperator.selector);
+        fcStorage.continuousCredit(start, end, units);
+    }
+
+    function testFuzzContinuousCredit(uint16 start, uint256 n, uint32 _units) public {
```
varunsrin, line 977: should we have a test for when end < start?

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1312 @@
...
+        fcStorage.setPrice(unitPrice);
+    }
+
+    function testFuzzSetUSDUnitPrice(uint256 unitPrice) public {
+        uint256 currentPrice = fcStorage.usdUnitPrice();
+
+        vm.expectEmit(false, false, false, true);
+        emit SetPrice(currentPrice, unitPrice);
+
+        vm.prank(admin);
```
varunsrin, line 1031: should we also test the treasurer here for thoroughness?

src/StorageRent.sol
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    /**
+     * @notice Change the maximum supply of storage units. Only callable by admin.
+     *
+     * @param max The new maximum supply of storage units.
+     */
+    function setMaxUnits(uint256 max) external onlyAdmin {
+        /* 1.6e7 = 64TB @ 4MB/unit */
+        if (max > 1.6e7) revert InvalidMaxUnits();
```
varunsrin, line 560: should we just set this to a much higher number (1e10) incase we ever decide to subdivide the units further?

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+        // Ensure Chainlink price is positive
+        newEthUsdPrice = bound(newEthUsdPrice, 1, type(int256).max);
+
+        // Set a new ETH/USD price
+        priceFeed.setPrice(newEthUsdPrice);
+
+        warp = bound(warp, 0, fcStorage.priceFeedCacheDuration());
+        vm.warp(block.timestamp + warp);
+
+        batchRentStorage(msgSender, ids, units);
```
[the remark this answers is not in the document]
horsefacts: Yeah, good call. We should do both—check the cached price and add an assertion inside the helper.

script/StorageRent.s.sol
varunsrin: oh you're right, they dont

test/StorageRent/StorageRent.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+        vm.prank(admin);
+        fcStorage.setMaxUnits(1.6e7);
+
+        uint256 length = _ids.length <= _units.length ? _ids.length : _units.length;
+        uint256[] memory ids = new uint256[](length);
+        for (uint256 i; i < length; ++i) {
+            ids[i] = _ids[i];
+        }
+
+        // Add an extra element to the units array
```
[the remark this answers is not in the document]
horsefacts: Yeah, `vm.assume` will throw away too many runs, so all the extra length munging in these is a workaround.

PR #238, opened 2023-06-30
fix(StorageRent): make tests faster by avoiding unnecessary event emit checks

varunsrin:
## Motivation

Makes tests 15x faster without sacrificing much coverage

## Change Summary

testFuzzContinuousCredit had ~ 512 runs generating 10,000 events each time, and each instance was being checked with `vm.expectEmit` which was very slow.  

The test was split into testContinuousCredit which checks that a single run with 20,000 calls emits events correctly and testFuzzContinuousCredit which does 512 runs with upto 1,000 calls for varying fids but only checks the final storage allocations and does not assert any events. 

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on adding a new test function and modifying an existing one in the `StorageRent` contract. 

### Detailed summary
- Added a new test function `testContinuousCredit` to simulate the initial seeding of the contract and check for emitted events.
- Modified the `testFuzzContinuousCredit` function to avoid checking for events and rely on `testContinuousCredit` instead.
- Added a new parameter `assertEvents` to the `continuousCredit` function to control whether to assert emitted events or not.

### DIFF
--- a/test/StorageRent/StorageRent.t.sol
+++ b/test/StorageRent/StorageRent.t.sol
@@ -1032,10 +1032,17 @@ contract StorageRentTest is StorageRentTestSuite {
         fcStorage.continuousCredit(start, end, units);
     }
 
+    function testContinuousCredit() public {
+        // Simulate the initial seeding of the contract and check that events are emitted.
+        continuousCredit(0, 20_000, 1, true);
+    }
+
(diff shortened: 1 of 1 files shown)


horsefacts (approved):
I like it!

PR #251, opened 2023-07-04
feat(StorageRent): add emergency price feed override.

horsefacts:
## Motivation

Our goal in using a price feed for storage payments is to maintain a unit price roughly around a fixed USD value, denominated in ETH. We don't need a perfectly up to date ETH/USD price for our use case, and anticipate caching the price feed price for 12-24 hours. 

There are many things that can go wrong when we depend on an external price feed, and many good defensive practices to protect against them: consulting multiple fallback feeds, triggering circuit breakers, etc. These are more important when up to date prices are mission critical. In our case, a simple emergency override (along with the feed validations we're already performing) is a good enough solution that protects against a reverting or malfunctioning feed.

If a price feed fails or malfunctions, we can fall back to a fixed price in storage and bypass external calls to the price feeds. 

## Change Summary

- Add a `fixedEthUsdPrice` parameter, settable by the admin role. This is not passed as a constructor parameter and initially zero.
- Use `fixedEthUsdPrice` as a price override when set. Returning early with this value disables the external calls to the data feeds altogether. Setting it back to zero re-enables feeds.
- Add admin setter functions/events for this parameter.

## Merge Checklist

- [x] This PR does not require changes to the [Farcaster protocol](http

---

## PR-Codex overview
This PR focuses on adding a fixed ETH/USD price as an emergency fallback in case of a price feed failure. 

### Detailed summary
- Added a `fixedEthUsdPrice` variable to the `StorageRent` contract.
- Added a `setFixedEthUsdPrice` function to set the fixed ETH/USD price.
- Emit an event when the fixed ETH/USD price is set.
- Updated the `price` function to use the fixed ETH/USD price if set.
- Added tests for renting with a fixed price and price feed failure.

> The following files were skipped due to too many changes: `test/StorageRent/StorageRent.t.sol`


test/StorageRent/StorageRent.t.sol
```diff
@@ -1103,18 +1231,81 @@ contract StorageRentTest is StorageRentTestSuite {
...
                               SET MAX UNITS
     //////////////////////////////////////////////////////////////*/
 
+    function testFuzzOnlyAdminCanSetMaxUnits(address caller, uint256 maxUnits) public {
+        vm.assume(caller != admin);
+
+        vm.prank(caller);
+        vm.expectRevert(StorageRent.NotAdmin.selector);
+        storageRent.setMaxUnits(maxUnits);
+    }
```
horsefacts, line 1307: This was in the wrong section, just moved it down here.

varunsrin:
@horsefacts lgtm, but needs conflicts resolved

Issue #255, opened 2023-07-06
feat: key registry contract


varunsrin:
**What is the feature you would like to implement?**
TBD

**Why is this feature important?**
TBD, new FIP needed

**Will the protocol spec need to be updated??**
Yes

PR #257, opened 2023-07-06
feat: add KeyRegistry

### LINKED ISSUE #258: feat: pass initial owner as constructor arg
**What is the feature you would like to implement?**

Our `Ownable2Step` contracts rely on the default `Ownable` behavior, which sets the owner to `msg.sender` at construction time. We should instead pass the initial owner as a constructor arg and transfer ownership from `msg.sender` in the constructor.

**Why is this feature important?**
If we deploy our contracts using a `CREATE2` factory, we need to transfer ownership from the factory contract address to the initial owner. I added a [fab contract](https://github.com/farcasterxyz/contracts/blob/main/script/helpers/IdRegistryFab.sol) to deploy and transfer atomically, but this is not necessary—we should simply transfer at construction time.

**Will the protocol spec need to be updated??**
No

**Additional context**
I intend to do this along with #257

horsefacts:
## Motivation

Implements [FIP: Onchain Signers](https://github.com/farcasterxyz/protocol/discussions/103), fixing #255. 

## Change Summary

- Adds a Key Registry contract as specified by the draft FIP
- Initializes contracts with an owner
- Updates default vscode settings to use solidity 0.8.19

## Merge Checklist

## Additional Context

Also closes #258

---

## PR-Codex overview
This PR focuses on adding an `owner` parameter to the constructors of `IdRegistry`, `FnameResolver`, `Bundler`, and `KeyRegistry` contracts. It also updates the test suites and harnesses accordingly.

### Detailed summary
- Added `owner` parameter to the constructors of `IdRegistry`, `FnameResolver`, `Bundler`, and `KeyRegistry`.
- Updated test suites and harnesses to pass the `owner` parameter.
- Updated the deployment script to use the new constructor syntax for `IdRegistry`.
- Updated the `FnameResolver` test to check the initial owner.
- Updated the `KeyRegistry` contract with new functionality and events.

> The following files were skipped due to too many changes: `src/KeyRegistry.sol`, `test/KeyRegistry/KeyRegistry.t.sol`


src/KeyRegistry.sol
```diff
@@ -0,0 +1,325 @@
...
+                uint256 fid = fids[i];
+                for (uint256 j = 0; j < keys[i].length; j++) {
+                    _remove(fid, 1, keys[i][j]);
+                }
+            }
+        }
+    }
+
+    function _register(uint256 fid, uint256 scope, bytes calldata key) internal {
+        Signer storage signer = signers[fid][scope][key];
```
sanjayprabhu, line 311: Hmm, so same signer can exist with multiple scopes? This is going to make the hub logic more complicated I think we should keep it to a single scope per key.

src/KeyRegistry.sol
```diff
@@ -0,0 +1,325 @@
...
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Register a public key to a fid/scope pair, setting the signer state to AUTHORIZED.
+     *
+     * @param fid   The fid associated with the key. Caller must own the provided fid.
+     * @param scope The key's numeric scope.
+     * @param key   Bytes of the signer's public key to authorize.
+     */
+    function register(uint256 fid, uint256 scope, bytes calldata key) external onlyFidOwner(fid) {
```
varunsrin, line 211: it might be useful to have an additional field called `metadata` which is optional and emitted with the event, containing additional detail about the signer, like a signature from the application that requested it

src/KeyRegistry.sol
```diff
@@ -0,0 +1,325 @@
...
+     * @param fid   The fid associated with the key. Caller must own the provided fid.
+     * @param scope The key's numeric scope.
+     * @param key   Bytes of the signer's public key to revoke.
+     */
+    function revoke(uint256 fid, uint256 scope, bytes calldata key) external onlyFidOwner(fid) {
+        Signer storage signer = signers[fid][scope][key];
+        if (signer.state != SignerState.AUTHORIZED && signer.state != SignerState.FROZEN) revert InvalidState();
+
+        signer.state = SignerState.REVOKED;
+        emit Revoke(fid, scope, key, key);
```
varunsrin, line 228: after thinking through the "freezing" logic, i think its a little too complicated to enforce right now on the hubs, and we should just get rid of the state to simplify this contract

src/KeyRegistry.sol
varunsrin: we probably also want a "register on behalf of" here

src/KeyRegistry.sol
```diff
@@ -0,0 +1,75 @@
...
+    error Unauthorized();
+
+    event Register(uint256 indexed fid, uint256 indexed scope, bytes indexed key);
+    event Revoke(uint256 indexed fid, uint256 indexed scope, bytes indexed key);
+    event Freeze(uint256 indexed fid, uint256 indexed scope, bytes indexed key);
+
+    IdRegistry public idRegistry;
+    uint40 public signersMigratedAt;
+
+    mapping(uint256 fid => mapping(uint256 scope => mapping(bytes key => Signer signer))) public signers;
```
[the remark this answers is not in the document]
varunsrin: also, we should move scope into the struct, otherwise the same key can have multiple scopes

src/KeyRegistry.sol
varunsrin: going to fix this

src/KeyRegistry.sol
varunsrin: tracked here: https://github.com/farcasterxyz/contracts/issues/263

src/KeyRegistry.sol
```diff
@@ -0,0 +1,299 @@
...
+    event Add(uint256 indexed fid, uint32 indexed scheme, bytes indexed key, bytes keyBytes, bytes metadata);
+
+    /**
+     * @dev Emit an event when an fid removes an added key.
+     *
+     * @param fid       The fid associated with the key.
+     * @param key       The key being registered. (indexed as hash)
+     * @param keyBytes  The bytes of the key being registered.
+     */
+    event Remove(uint256 indexed fid, bytes indexed key, bytes keyBytes);
```
TimDaub, line 72: Why does Remove have `bytes indexed key` and `bytes keyBytes` if both are set to `key` in its only usage: https://github.com/farcasterxyz/contracts/pull/257/files#diff-ffa5866d21942cfa6a42043498fbb08d5ef20a0d5d97ba107f6d94fb33713e59R207 ?

src/KeyRegistry.sol
```diff
@@ -0,0 +1,299 @@
...
+        if (keyData.state != KeyState.NULL) revert InvalidState();
+
+        keyData.state = KeyState.ADDED;
+        keyData.scheme = scheme;
+        emit Add(fid, scheme, key, key, metadata);
+    }
+
+    function _reset(uint256 fid, bytes calldata key) internal {
+        KeyData storage keyData = keys[fid][key];
+        if (keyData.state != KeyState.ADDED) revert InvalidState();
```
sanjayprabhu, line 293: Why have this check? Could be more useful to be able to clear the map regardless of state.

src/KeyRegistry.sol
```diff
@@ -0,0 +1,299 @@
...
+    /**
+     * @dev Emit an event when an admin or fid adds a new key.
+     *
+     * @param fid       The fid associated with the key.
+     * @param scheme    The type of the key.
+     * @param key       The key being registered. (indexed as hash)
+     * @param keyBytes  The bytes of the key being registered.
+     * @param metadata  Metadata about the key.
+     */
+    event Add(uint256 indexed fid, uint32 indexed scheme, bytes indexed key, bytes keyBytes, bytes metadata);
```
sanjayprabhu, line 63: Aren't key and keyBytes always the same? Why do we need to repeat them?

src/KeyRegistry.sol
```diff
@@ -0,0 +1,299 @@
...
+     * @param fid       The fid associated with the key.
+     * @param key       The key being reset. (indexed as hash)
+     * @param keyBytes  The bytes of the key being registered.
+     */
+    event AdminReset(uint256 indexed fid, bytes indexed key, bytes keyBytes);
+
+    /**
+     * @dev Emit an event when the admin calls migrateKeys.
+     */
+    event Migrated();
```
sanjayprabhu, line 86: We should include `keysMigratedAt` here to make it easy for the hubs to track the timestamp.

src/KeyRegistry.sol
varunsrin: cc @horsefacts for context on this

src/KeyRegistry.sol
varunsrin: I went back and forth on this. Rationale is that this will help us prevent migration errors where we might incorrectly seed an array. I don't hold this strongly, and we might change our minds on this after doing the testnet migration.

src/KeyRegistry.sol
```diff
@@ -0,0 +1,299 @@
...
+
+    /**
+     * @dev Mapping of fid to a key to the key's metadata.
+     *
+     * @custom:param fid       The fid associated with the key.
+     * @custom:param key       Bytes of the key.
+     * @custom:param data      Struct with the state and key type. In the initial migration
+     *                         all keys will have data.scheme == 1.
+     */
+    mapping(uint256 fid => mapping(bytes key => KeyData data)) public keys;
```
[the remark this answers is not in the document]
varunsrin: hmm, good point on L2 event history. i'll open a ticket to figure this out and resolve.

src/KeyRegistry.sol
TimDaub: I was also initially wondering but then I read that solidity automatically keccak hashes byte arrays that are above a certain size in an event. TIL

Issue #256, opened 2023-07-06
bug: griefing vectors with simple recovery


varunsrin:
### **What is the bug?**

Simplification of the recovery system combined with the ability to "register to" any address has introduced new vectors for griefing registrations for users. 

_Originally reported by @horsefacts_ 

### **How can it be reproduced?**

**1. Registration Stealing**

- Alice calls `register(alice, bob)` to register an fid
- Mallory sees Alice's tx, frontruns it, and either calls `register(alice, mallory)` or `transfer()`'s or `recover()`'s a previously registered id into this address which has mallory as its recovery address
- Alice's registration tx fails since she now owns Mallory's malicious fid.
- Alice might not notice, and validate other things like her ENS Name
- Mallory can `recover()` Alice's fid to her address at a later date (assuming alice hasn't noticed) 

**2. Account Griefing**

- Mallory `registers()` an fid, does a bunch of bad stuff
- She then `transfer()`'s or `recover()`'s it to bob to associate it with him
- Bob now has a bad reputation and must spend gas to delete the address 
- Mallory could further grief bob by sandwiching bob's attempt to transfer the fid with two `recover()`s or a `recover()` and a `transfer()` 
 

**3. Registration DoS**

- Alice tries to call `register()` on an address
- Mallory sandwiches the call by frontrunning it with a `register(alice, mallory)` or a `register(mallory, mallory)` which is accompanied by a `transfer(mallory, alice)` or a `recover(mallory, alice)`
- This blocks alice's `register()` transaction
- Mallory completes the sandwich by then calling `recover(alice, mallory)` to move the fid out before alice can remove her as the recovery address.

varunsrin:
from @horsefacts 

> One solution that preserves the ability to register/transfer on behalf of another might be to require a signed message from the to address in register, trustedRegister, transfer and maybe even recover. If a client is registering an fid on behalf of a user, this has no gas cost to the user, but they do have to sign a message at registration time.

varunsrin:
Some initial thoughts: 

- Registration stealing is preventable if farcaster wallet apps are vigilant. Their transaction will fail and they can check the chain to see that the registered fid has a recovery address and prompt the user to accept it. If they write lazy checks like just seeing if the address has an fid, they might miss this entirely (possible, since apps may use the fid existence check to verify your ownership).

- Registration stealing can be eliminated only by requiring that register and transfer provide a signed proof from the address that the fid is going to end up in. 

[...]

PR #261, opened 2023-07-07
feat(IdRegistry): require sigs when registering, transferring or recovering to another address

horsefacts:
## Motivation

#256 

## Change Summary

- Add signature verification to `register` which is now called `registerFor`.
- Added a new method `register` which only registers to the caller
- Added signature verification to `transfer` and `recover`
- Removed meta-txns which are no longer necessary

Calling `registerFor`, `transfer` or `recover` now takes two additional args: a `uint256 deadline` timestamp and a `bytes calldata` EIP712 signature with typehash `Register(address to,address recovery,uint256 nonce,uint256 deadline)` that must recover to the `to` address. This prevents registration frontrunning and griefing but allows anyone to register on behalf of another by providing a signature from the recipient of the fid.

Tests are slightly more complicated now that we need to generate signatures and have changed in a few ways:
- Instead of fuzzing address, we sometimes fuzz private keys, e.g. `address bob` -> `uint256 bobPk`, since we need the key to generate a signature. You can get the corresponding address with `vm.addr(pk)`. 
- Not all `uint256` values are valid private keys, only those in the lower half of the secp256k1 curve order. So there's a `_boundPk()` helper that will bound the fuzzed `uint256` to the right range.
- Signatures require a deadline to prevent replay, so this is now an additional fuzz parameter in a bunch of places.
- Similarly, there's a `_boundDeadline` helper that converts a fuzzed `uint40` timestamp to a valid `uint256` deadline in the future.
- Finally, there is a `_signRegisterFor` helper that signs a register message for the given private key, `to`, `recovery`, and `deadline`.

## Merge Checklist

## Additional Context

---

### Detailed summary
- Deleted `test/IdRegistry/IdRegistry.metatx.t.sol.env.example` file.
- Removed `ID_REGISTRY_TRUSTED_FORWARDER_ADDRESS` variable.
- Updated `IdRegistryFab` constructor to only take `initialOwner` parameter.
- Updated `register` function in `Bundler.sol` to include `deadline` and `sig` parameters.
- Renamed `testGasRegister` function to `testGasRegisterWithSig` in `Bundler.gas.t.sol`.
- Updated `run` function in `IdRegistry.s.sol` to remove `trustedForwarder` parameter in `IdRegistryFab` constructor.
- Added `hashTypedDataV4`, `registerTypehash`, and `transferTypehash` functions in `IdRegistryHarness.sol`.
- Updated `IdRegistryOwnerTest` contract to remove references to `FORWARDER` constant.
- Updated `setUp` function in `IdRegistryTestSuite.sol` to remove `FORWARDER` parameter in `IdRegistryHarness` constructor.
- Added `_registerFor`, `_registerForWithRecovery`, `_boundPk`, and `_boundDeadline` functions in `IdRegistryTestSuite.sol`.

> The following files were skipped due to too many changes: `test/IdRegistry/IdRegistryTestSuite.sol`, `test/Bundler/Bundler.t.sol`, `src/IdRegistry.sol`, `test/IdRegistry/IdRegistry.t.sol`


src/Bundler.sol
```diff
@@ -89,8 +89,14 @@ contract Bundler is Ownable2Step {
      * @param recovery     Address that is allowed to perform a recovery
      * @param storageUnits Number of storage units to rent
      */
-    function register(address to, address recovery, uint256 storageUnits) external payable {
-        uint256 fid = idRegistry.register(to, recovery);
+    function register(
```
varunsrin, line 92: needs to be replicated for recovery and transfer to meet all the requirements

PR #275, opened 2023-07-18
test: add symbolic tests for IdRegistry and KeyRegistry invariants

daejunpark:
## Motivation

The purpose of this PR is to illustrate and discuss the use of symbolic tests for verifying contract invariants. Symbolic tests offer several advantages over random testing (such as foundry fuzzing and invariant testing), and can be highly effective when used properly. The tests provided in this (draft) PR is not exhaustive, but should serve as a pointer to provide the idea of symbolic tests and help further discussion.

## Change Summary

- Install the halmos-cheatcode library. (This will be re-installed as submodule after the library is released next week.)
- Add symbolic tests for IdRegistry and KeyRegistry.
- Update CI to run the symbolic tests with Halmos.

## Merge Checklist

---

### Focus of this PR:
This PR focuses on adding symbolic tests for the `IdRegistry` and `KeyRegistry` contracts.

### Detailed summary:
- Added symbolic tests for the `IdRegistry` contract.
- Added symbolic tests for the `KeyRegistry` contract.
- Added imports for necessary contracts and interfaces.
- Added setup functions to initialize contract states.
- Added helper functions for generating valid calldata.
- Added assertions to verify invariants and state transitions.
- Added SPDX license identifiers to the test files.

> The following files were skipped due to too many changes: `test/KeyRegistry/KeyRegistry.symbolic.t.sol`


horsefacts:
Thanks for all the setup here, @daejunpark! Halmos has come a long way and this is a very useful example.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,26 @@ jobs:
...
+
+      - name: Install foundry
+        uses: foundry-rs/foundry-toolchain@v1
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install halmos
+        run: pip install git+https://github.com/a16z/halmos
```
varunsrin, line 59: self: pin to specific version

test/IdRegistry/IdRegistry.st.sol
```diff
@@ -0,0 +1,192 @@
...
+        // If a new fid is registered, ensure that:
+        // - IdRegistry must not be paused.
+        // - idCounter must increase by 1.
+        // - The new fid must not be registered for an existing fid owner.
+        // - The existing fids must be preserved.
+        if (newIdCounter > oldIdCounter) {
+            assert(oldPaused == false);
+            assert(newIdCounter - oldIdCounter == 1);
+            assert(newIdX == oldIdX || oldIdX == 0);
+            assert(newIdX == oldIdX || newIdY == oldIdY);
```
varunsrin, line 93: I found these last two lines hard to reason about, and rewrote them as follows. From the POV of symbolic verification, is there any difference between what we are asserting here?

```
            if (newIdX != oldIdX) {
                assert(oldIdX == 0 && newIdX == newIdCounter);
                assert(oldIdY == newIdY);
            } else if (newIdY != oldIdY) {
                assert(oldIdY == 0 && newIdY == newIdCounter);
                assert(oldIdX == newIdX);
            } else {
                assert(oldIdX == newIdX && oldIdY == newIdY);
            }
```

test/IdRegistry/IdRegistry.st.sol
```diff
@@ -0,0 +1,192 @@
...
+                svm.createUint256("deadline"),
+                svm.createBytes(65, "sig")
+            );
+        } else {
+            args = svm.createBytes(1024, "data");
+        }
+        return abi.encodePacked(selector, args);
+    }
+
+    function check_transfer(address caller, address to, address other) public {
```
varunsrin, line 122: we test transfer in two ways - via this specific test for transfer, and more generically in check_invariant, though we do not assert much there. 

we test register only in one way - in the general check invariant, where we assert a lot of specific stuff. 

is there a rationale for this design? or is this something that could be standardizes where we have a general invariant assertion test and specific ones for each function including register?

test/IdRegistry/IdRegistry.st.sol
varunsrin: very helpful thanks.

1. Agree that the last branch is unnecessary, and can be removed. 
2. What are the implications of logical duality here? Does it imply that this code is testing different things than the original logic you specified?

test/IdRegistry/IdRegistry.symbolic.t.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        assert(idRegistry.getRecoveryOf(1) == address(0x2001));
+        assert(idRegistry.getRecoveryOf(2) == address(0x2002));
+
+        // Create symbolic addresses
+        x = svm.createAddress("x");
+        y = svm.createAddress("y");
+    }
+
+    function verifyInvariants(bytes4 selector, address caller) public {
```
daejunpark, line 39: note that the default prefix for halmos test functions is `check_`. if you want to use a different prefix, then you need to provide an additional option `--function <prefix>` to halmos, which is `--function verify` in this case. otherwise these functions will be omitted from testing, as you can see in the last ci run:
https://github.com/farcasterxyz/contracts/actions/runs/5671611054/job/15369109292

horsefacts:
Thanks for all your help here and on tg, @daejunpark and @karmacoma-eth! I'm working on updating the tests here to account for some contract changes and have bumped into a counterexample that I don't understand.

I've updated the tests to account for a few changes to the key registry:
- Add/remove no longer take a leading `fid` argument. 
- New `addFor`, `trustedAdd`, and `removeFor` functions.
- Renamed `pauseRegistration` to `pause`.

Since `addFor`, `trustedAdd`, and `removeFor` all take at least one `bytes` argument, I added them to `mk_calldata` following the existing example. I also introduced new symbolic values for their arguments: `address addr`, `uint256 deadline`, and `bytes sig`. 

I updated the invariants to account for these new functions. Since there are now multiple add/remove functions, there's now an extra check inside each post-state to detect which function was called and make any function-specific assertions. For example, before:

```solidity
            if (newStateX == KeyRegistry.KeyState.REMOVED) {
                // For a transition to REMOVED, ensure that:
                // - The previous state must be ADD.
                // - The transition can only be made by remove(), where:
                //   - It must be called by the owner of fid x.
                assert(oldStateX == KeyRegistry.KeyState.ADDED);
                assert(selector == keyRegistry.remove.selector);
                assert(oldCallerId == x);
            } 
``` 

And after:

```solidity
            if (newStateX == KeyRegistry.KeyState.REMOVED) {
                // For a transition to REMOVED, ensure that:
                // - The previous state must be ADD.
                assert(oldStateX == KeyRegistry.KeyState.ADDED);
                // - The transition can only be made by remove() or removeFor()
                if (selector == keyRegistry.remove.selector) {
                    //   - remove() must be called by the owner of fid x.
                    assert(oldCallerId == x);
                } else if (selector == keyRegistry.removeFor.selector) {
                  // No assertion here yet
                } else {
                    assert(false);
                }
            }
```

However, introducing the `assert(oldCallerId == x)` assertion [here](https://github.com/farcasterxyz/contracts/blob/ae042943fbe27ebd33b65c70b7289275fb23f956/test/KeyRegistry/KeyRegistry.symbolic.t.sol#L120) fails and produces a counterexample: 

```
Counterexample:
    halmos_disableTrustedOnly?_bool_07 = 0x0
    halmos_gracePeriod_uint24_01 = 0x640001
    halmos_key1_bytes_02 = 0x0
    halmos_key2_bytes_03 = 0x1a
    halmos_key_bytes_13 = 0x1a
    halmos_migrateKeys?_bool_06 = 0x1
    halmos_pause?_bool_08 = 0x0
    halmos_timestamp2_uint64_09 = 0x0
    halmos_x_uint256_04 = 0x2
    halmos_xkey_bytes_05 = 0x1a
    p_caller_address = 0x8000000000000000000000000000000000000000
    p_selector_bytes4 = 0x58edef4c00000000000000000000000000000000000000000000000000000000
```

The function selector here is `remove(bytes)`, and the FID removed is 2, which we assign to `address(0x1002)` in [setup](https://github.com/farcasterxyz/contracts/blob/ae042943fbe27ebd33b65c70b7289275fb23f956/test/KeyRegistry/KeyRegistry.symbolic.t.sol#L36) and use to register a key [here](https://github.com/farcasterxyz/contracts/blob/ae042943fbe27ebd33b65c70b7289275fb23f956/test/KeyRegistry/KeyRegistry.symbolic.t.sol#L68).

 If I'm reading the failure right, it indicates that `address(0x8000000000000000000000000000000000000000)` was able to call `remove` and transition FID 2 to `REMOVED`. But when I run this as a concrete test with the values from the counterexample, it reverts, since the caller is not the FID owner. This is what I'd expect intuitively, and what actually happens when I concretize this example.
 
 Am I interpreting the counterexample correctly? Anything else look suspicious here?

karmacoma-eth:
ngl this is a chonky test, this is why we need https://github.com/a16z/halmos/issues/130

daejunpark:
thanks for updating the tests!  let me take a look at the failure.

horsefacts:
> ngl this is a chonky test

Ha, yes, and I'm afraid I made it chonkier. 

Not sure either of these are relevant, but I noticed two other things here:
- We create two `key2` symbolic variables, one on [L65](https://github.com/farcasterxyz/contracts/blob/ae042943fbe27ebd33b65c70b7289275fb23f956/test/KeyRegistry/KeyRegistry.symbolic.t.sol#L65) and one inside `mk_calldata` on [L180](https://github.com/farcasterxyz/contracts/blob/ae042943fbe27ebd33b65c70b7289275fb23f956/test/KeyRegistry/KeyRegistry.symbolic.t.sol#L180).
- In all the counterexamples, `key2`, `key` and `xkey` look like they are equal.

daejunpark:
it turned out that a known halmos bug was triggered by this test. i've temporarily adjusted the test to avoid the bug and will fix it quickly. will get back to this soon.

daejunpark:
> Ha, yes, and I'm afraid I made it chonkier.

No, this test is one of the good uses of halmos.  What is now missing from halmos is the debugging info, such as the trace info mentioned by Karma.

daejunpark:
> it turned out that a known halmos bug was triggered by this test. i've temporarily adjusted the test to avoid the bug and will fix it quickly. will get back to this soon.

i fixed the halmos issue, and reverted the temporary adjustment to the test. i also rebased the pr on the main branch, and updated the test for the new interface. it should work again now. please let me know if you have any further questions.

daejunpark:
the tests passed locally, but seems too heavy for the github ci. let me try optimizing them.

horsefacts:
Splitting this into #397 and #398 so we can merge the ID registry tests separately. Will close this and move conversation to #398.

PR #279, opened 2023-07-24
fix(KeyRegistry): prevent add for zero fid

### LINKED ISSUE #272: review: registration of key on fid 0
The current KeyRegistry.add() function does not prevent regular users from adding a key for fid 0.

It is recommended to make add() to revert when fid == 0, especially if such a key for fid 0 could potentially be exploited in the offchain logic (or in the future).

horsefacts:
## Motivation

Prevent key registrations to the zero fid. See #272.

## Change Summary

Revert in `_add` if `fid` is zero.

## Merge Checklist

## Additional Context

Close #272 

---

### Detailed summary

This PR focuses on adding a new function `testFuzzRegisterRevertsFidZero` to the `KeyRegistry` contract. Notable changes include:

- Addition of the `testFuzzRegisterRevertsFidZero` function in `KeyRegistry.t.sol`
- Implementation of the `_add` function in `KeyRegistry.sol` with a check for unauthorized access

### DIFF
--- a/src/KeyRegistry.sol
+++ b/src/KeyRegistry.sol
@@ -327,6 +327,7 @@ contract KeyRegistry is Ownable2Step {
     //////////////////////////////////////////////////////////////*/
 
     function _add(uint256 fid, uint32 scheme, bytes calldata key, bytes calldata metadata) internal {
+        if (fid == 0) revert Unauthorized();
         KeyData storage keyData = keys[fid][key];
         if (keyData.state != KeyState.NULL) revert InvalidState();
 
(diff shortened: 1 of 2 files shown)


varunsrin:
PR looks good, but per discussion I don't know if this is really a bug we care about? If someone assigns a signer to the 0th fid, that's not an issue since no one can own the 0th fid. Avoiding this check will also save a small amount of gas on every signer add, so I'm in favor of reverting it. 

The only thing we'd need to make sure is that hubs interpret this in a reasonable way cc @sanjayprabhu

horsefacts:
I think I agree. The one edge case this does catch is if we accidentally add a key for FID 0 during migration. If we're careful to avoid that, I think we can omit this.

sanjayprabhu:
Yes, this should be fine. The event will be rejected by the hubs since fid 0 is not registered.

PR #296, opened 2023-07-31
perf(TransferHelper): remove unnecessary checks and use yul for native token transfer

fiveoutofnine:
## Motivation

Removes an unnecessary check for balance before transferring native tokens (the call would return false if insufficient balance anyway).

## Change Summary

- Remove `InsufficientFunds` check
- Transfer ETH with inline assembly (similar implementation as [**solmate/utils/SafeTransferLib.sol**](https://github.com/transmissions11/solmate/blob/bfc9c25865a274a7827fea5abf6e4fb64fc64e6c/src/utils/SafeTransferLib.sol#L15))

## Merge Checklist

## Additional Context

Not a complex change, but the code should function identical to [**solmate/utils/SafeTransferLib.sol**](https://github.com/transmissions11/solmate/blob/bfc9c25865a274a7827fea5abf6e4fb64fc64e6c/src/utils/SafeTransferLib.sol#L15), which is pretty [**well-tested**](https://github.com/transmissions11/solmate/blob/bfc9c25865a274a7827fea5abf6e4fb64fc64e6c/src/test/SafeTransferLib.t.sol#L18).

### DIFF
--- a/src/lib/TransferHelper.sol
+++ b/src/lib/TransferHelper.sol
@@ -5,24 +5,18 @@ library TransferHelper {
     /// @dev Revert when a native token transfer fails.
     error CallFailed();
 
-    /// @dev Revert when there are not enough funds for a native token transfer.
-    error InsufficientFunds();
-
     /**
      * @dev Native token transfer helper.
      */
     function sendNative(address to, uint256 amount) internal {
-        if (address(this).balance < amount) revert InsufficientFunds();
+        bool success;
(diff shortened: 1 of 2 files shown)


fiveoutofnine:
fyi `forge snapshot --match-contract GasUsage` yields no changes, but I think that's because `TransferHelper.sendNative` is never called in any of those gas usage tests

PR #297, opened 2023-07-31
feat(IdRegistry): external view to verify an fid's signature

Hmac512:
## Motivation

This PR adds an external view to the IdRegistry contract such that a fid's signature can be verified. Having a single way to verify a fid signature will make the developer experience easier since the registry has 1271 signature support.

https://t.me/farcasterdevchat/6142 (original telegram thread on this topic)

## Change Summary

This PR:

1. Adds an external view to IdRegistry that verifies a fid signature by checking if the signer owns the fid, and if the signature is correct.
2. Adds tests to make sure the view works as exptected.

## Merge Checklist

## Additional Context

This is a pretty simple change. It only affects external calls. None of the internal calls  (registering, transferring etc) are affected.

---

### Detailed summary
- Added a new internal function `_signDigest` in `IdRegistryTestSuite.sol` to sign a digest using a private key.
- Added a new external view function `verifyFidSignature` in `IdRegistry.sol` to verify the signature of a custody address for a given fid and digest.
- Added new test functions in `IdRegistry.t.sol` to test the `verifyFidSignature` function.


Hmac512:
I can't change the labels after creating the PR :(

src/IdRegistry.sol
```diff
@@ -318,4 +317,26 @@ contract IdRegistry is TrustedCaller, Signatures, Pausable, EIP712, Nonces {
...
+    /**
+     * @notice Verify a signature from a fid owner.
+     *
+     * @param signer   The address to check the signature of.
+     * @param fid      The fid to check the signature of.
+     * @param digest   The dgiest that was signed.
+     * @param sig      The signature to check.
+     */
+    function verifyFidSignature(
+        address signer,
```
varunsrin, line 334: should call this `custody address` to match terminology elsewhere

src/IdRegistry.sol
```diff
@@ -318,4 +317,26 @@ contract IdRegistry is TrustedCaller, Signatures, Pausable, EIP712, Nonces {
...
+    /*//////////////////////////////////////////////////////////////
+                             EXTERNAL VIEW HELPERS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Verify a signature from a fid owner.
+     *
+     * @param signer   The address to check the signature of.
+     * @param fid      The fid to check the signature of.
+     * @param digest   The dgiest that was signed.
```
varunsrin, line 330: typo: digest

src/lib/Signatures.sol
```diff
@@ -18,6 +18,10 @@ abstract contract Signatures {
                      SIGNATURE VERIFICATION HELPERS
     //////////////////////////////////////////////////////////////*/
 
+    function _verifySigPlain(bytes32 digest, address signer, bytes memory sig) internal view returns (bool is_valid) {
```
varunsrin, line 21: do we need this helper? it's easier to understand whats going on if we just call it inline in the verifyFidSignature method.

src/IdRegistry.sol
```diff
@@ -318,4 +317,26 @@ contract IdRegistry is TrustedCaller, Signatures, Pausable, EIP712, Nonces {
             sig
         );
     }
+
+    /*//////////////////////////////////////////////////////////////
+                             EXTERNAL VIEW HELPERS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Verify a signature from a fid owner.
```
varunsrin, line 326: nit: Verify that a signature was produced by the custody address that owns an  fid.

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,74 @@ contract IdRegistryTest is IdRegistryTestSuite {
         assertEq(idRegistry.getRecoveryOf(1), address(0));
     }
 
+    function testFuzzVerifyFidSignature(
```
varunsrin, line 303: nit add comment headers to separate and make it easier to read, like: 

```
    /*//////////////////////////////////////////////////////////////
                         TRUSTED REGISTER TESTS
```

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,74 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        address registrar,
+        uint256 recipientPk,
+        address recovery,
+        uint40 _deadline
```
varunsrin, line 307: we don't need to fuzz `registrar`, `_deadline` or `recovery` here and we should be fuzzing `digest` in addition to `recipientPk`

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,74 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        assertEq(idRegistry.getIdCounter(), 1);
+        assertEq(idRegistry.idOf(recipient), 1);
+        assertEq(idRegistry.getRecoveryOf(1), recovery);
+
+        bytes32 digest = keccak256(bytes("message"));
+        bytes memory msg_sig = _signMsg(recipientPk, digest);
+        assertEq(idRegistry.verifyFidSignature(recipient, 1, digest, msg_sig), true);
+    }
+
+    function testFuzzCannotVerifyFidSignature(
```
varunsrin, line 336: same comments about picking the right fuzz params

varunsrin (changes requested):
Left some comments on things that should be changed - core implementation is fine, but we can tighten up the tests a bit.

src/lib/Signatures.sol
Hmac512: I thought I'd add it if the same sort of external view would be used somewhere else.

Removed.

src/IdRegistry.sol
Hmac512: Changed

test/IdRegistry/IdRegistry.t.sol
Hmac512: Added better seperator

test/IdRegistry/IdRegistry.t.sol
Hmac512: Cleaned up the tests

Hmac512:
> Left some comments on things that should be changed - core implementation is fine, but we can tighten up the tests a bit.

Resolved comments, and cleaned up tests.

src/IdRegistry.sol
varunsrin: I meant that the `signer` param should be called `custodyAddress` instead. 

we should keep `verifyFidSignature` as is

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
         assertEq(idRegistry.getRecoveryOf(1), address(0));
     }
 
+    /*//////////////////////////////////////////////////////////////
+                         VERIFY FID CUSTODY ADDRESS TESTS
+    //////////////////////////////////////////////////////////////*/
+
+    function testFuzzVerifyFidCustody(uint256 recipientPk, bytes32 digest) public {
+        uint256 deadline = _boundDeadline(0);
```
varunsrin, line 308: nit: do we need to bound deadline? could just use some hardcoded value here.

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        assertEq(idRegistry.verifyFidCustodyAddress(recipient, 1, digest, msg_sig), true);
+    }
+
+    function testFuzzCannotVerifyFidCustodyIfBadDigest(
+        uint256 recipientPk,
+        bytes32 digest,
+        bytes32 bad_digest
+    ) public {
+        assert(digest != bad_digest);
+        uint256 deadline = _boundDeadline(0);
```
varunsrin, line 328: nit: same comment about bounding deadline

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        recipientPk = _boundPk(recipientPk);
+        address recipient = vm.addr(recipientPk);
+        bytes memory sig = _signRegister(recipientPk, recipient, address(0), deadline);
+
+        vm.prank(owner);
+        idRegistry.disableTrustedOnly();
+        idRegistry.registerFor(recipient, address(0), deadline, sig);
+        assertEq(idRegistry.getIdCounter(), 1);
+
+        bytes memory msg_sig = _signMsg(recipientPk, digest);
```
varunsrin, line 318: nit: camel case solidity variable names

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
```
varunsrin, line 338: nit: camel case solidity variable names

test/IdRegistry/IdRegistryTestSuite.sol
```diff
@@ -82,4 +82,10 @@ abstract contract IdRegistryTestSuite is TestSuiteSetup {
         signature = abi.encodePacked(r, s, v);
         assertEq(signature.length, 65);
     }
+
+    function _signMsg(uint256 pk, bytes32 digest) internal returns (bytes memory signature) {
```
varunsrin, line 86: nit: call this `_signDigest` to line up with the params or just `_sign`

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        assertEq(idRegistry.verifyFidCustodyAddress(recipient, 1, digest, msg_sig), true);
+    }
+
+    function testFuzzCannotVerifyFidCustodyIfBadDigest(
+        uint256 recipientPk,
+        bytes32 digest,
+        bytes32 bad_digest
```
varunsrin, line 325: camelCase variable names

src/IdRegistry.sol
```diff
@@ -318,4 +318,25 @@ contract IdRegistry is TrustedCaller, Signatures, Pausable, EIP712, Nonces {
...
+     * @param digest   The digest that was signed.
+     * @param sig      The signature to check.
+     */
+    function verifyFidCustodyAddress(
+        address signer,
+        uint256 fid,
+        bytes32 digest,
+        bytes calldata sig
+    ) external view returns (bool is_valid) {
+        is_valid = idOf[signer] == fid && SignatureChecker.isValidSignatureNow(signer, digest, sig);
```
varunsrin, line 340: should we just look up the address here instead of asking the user to provide it? is there a benefit to asking?

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,45 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        vm.prank(owner);
+        idRegistry.disableTrustedOnly();
+        idRegistry.registerFor(recipient, address(0), deadline, sig);
+        assertEq(idRegistry.getIdCounter(), 1);
+
+        bytes memory msg_sig = _signMsg(recipientPk, digest);
+        assertEq(idRegistry.verifyFidCustodyAddress(recipient, 1, digest, msg_sig), true);
+    }
+
+    function testFuzzCannotVerifyFidCustodyIfBadDigest(
```
varunsrin, line 322: we should also have tests that verify failure if the fid is specified incorrectly, and potentially the custody address if we decide to retain tht parameter.

varunsrin (changes requested):
Great work so far, just had a couple more followups

src/IdRegistry.sol
Hmac512: https://github.com/farcasterxyz/contracts/blob/ccb80bb8a7edaa713552d4f81dc298856c3a1b4c/src/IdRegistry.sol#L102-L107

There is no mapping from fid to owner address.

There is only owner->fid, and fid->recoveryAddress.


```
    function _unsafeRegister(address to, address recovery) internal whenNotPaused returns (uint256 fid) {
        /* Revert if the target(to) has an fid */
        if (idOf[to] != 0) revert HasId();

        /* Safety: idCounter won't realistically overflow. */
        unchecked {
            /* Incrementing before assignment ensures that no one gets the 0 fid. */
            fid = ++idCounter;
        }

        /* Perf: Save 29 gas by avoiding to == address(0) check since 0x0 can only register 1 fid */
        idOf[to] = fid;
        recoveryOf[fid] = recovery;
    }
```

The contract assumes an address can only have 1 fid. There is no way to resolve a fid to it's owner address from the contract.

Would require a pretty significant change to the contract to add fid->owner resolution. You probably want to just make it a NFT contract with a 1 nft per address limitation at that point.

Hmac512:
Made most changes requested.

Pretty sure adding fid->owner resolution is out of scope for this PR.

test/IdRegistry/IdRegistry.t.sol
Hmac512: I added 2 extra tests.

src/IdRegistry.sol
Hmac512: One alternative is to not as for a fid, and return the fid is sig is valid or throw an error. 

I think this alternative should be a separate view func though.

The main point of the `verifyFidSignature` is to make the developer experience easier if anyone needs to verify a fid signature off-chain. This is important considering the erc-1271 support that was added.


We could also add a `recoverFid` that doesn't take a fid (still need custodyAddress). This could also be useful to have.

test/IdRegistry/IdRegistry.t.sol
```diff
@@ -300,6 +300,75 @@ contract IdRegistryTest is IdRegistryTestSuite {
...
+        assertEq(idRegistry.getIdCounter(), 1);
+
+        bytes memory msgSig = _signDigest(recipientPk, digest);
+        assertEq(idRegistry.verifyFidSignature(recipient, 2, digest, msgSig), false);
+    }
+
+    function testFuzzCannotVerifyFidSignatureIfBadCustodyAddress(
+        uint256 recipientPk,
+        bytes32 digest,
+        address badCustodyAddress
```
varunsrin, line 357: is it possible that badCustodyAddress collides with the public key derived from recipientPk? 

if so, we should probably vm.assume that case out.

Hmac512: Still getting used to Foundry. Much better with Hardhat.

I added vm.assume where it makes sense in the added tests.

varunsrin (approved):

Issue #295, opened 2023-07-31
bug: idregistry state cannot be frozen for a future upgrade


varunsrin:
**What is the bug?**
There is no way to freeze the state of IdRegistry, which is important for a safe upgrade in case a bug is discovered in the future. The closest function is pause, which does not prevent transfers or recoveries. Consider extending pause to freeze all state or add a new one-way method like deprecate which freezes the contract entirely.

horsefacts:
@varunsrin I think this was closed by #301 Anything remaining?

PR #301, opened 2023-08-01
feat(IdRegistry): fail transfer, recovery, changeRecovery when paused

varunsrin:
## Motivation

If IdRegistry needs to be redeployed it is useful to have a way to freeze the old contract so that all state can be copied over to the new one. The `pause` function previously allowed transfers, recoveries and changing the recovery address which meant that it could not be used to guarantee that state would not change. 

## Change Summary

Revert transfer, changeRecoveryAddress and recover if the contract is paused.

## Merge Checklist


horsefacts (approved):
Close #295

PR #308, opened 2023-08-02
feat(StorageRegistry): price bounds, testing and documentation improvements

varunsrin:
## Motivation

Fixed several issues that came up in the review of the Storage Registry contracts.

## Change Summary

- Add configurable minimum and maximum values for exchange rates returned by Chainlink.
- Do not depend on deprecated `answeredInRound` property.
- Increase test coverage by adding states or more assertions. 
- Rename functions and variables for clarity.

## Merge Checklist

---

### Detailed summary
- Added two new constants `INITIAL_PRICE_FEED_MIN_ANSWER` and `INITIAL_PRICE_FEED_MAX_ANSWER` in `StorageRegistryTestSuite.sol`
- Added error types `InvalidMinAnswer` and `InvalidMaxAnswer` in `StorageRegistry.sol`
- Added events `SetMinAnswer` and `SetMaxAnswer` in `StorageRegistry.sol`
- Added storage variables `priceFeedMinAnswer` and `priceFeedMaxAnswer` in `StorageRegistry.sol`
- Added functions `setMinAnswer` and `setMaxAnswer` in `StorageRegistry.sol`
- Modified function `_refreshPrice` in `StorageRegistry.sol` to validate price feed answer against the min and max bounds

> The following files were skipped due to too many changes: `test/StorageRegistry/StorageRegistry.t.sol`


varunsrin:
A few notes for discussion: 

1. The [Data Feeds API Reference](https://docs.chain.link/data-feeds/api-reference) say that answeredInRound is deprecated, but we use it in multiple places as a check. 
2. For StorageRent Uptime Feed tests, should we be fuzzing some of the values that we're mocking to get more thorough coverage?
3. Integration tests like testFuzzUptimeFeedFailure could be separated from unit tests to make it easier to read. 
4. Should we be unit testing methods like `_refreshPrice` instead of external methods like `refreshPrice`?  
5. StorageRent has an invalid price check that if the ETH-USD rate goes to 0 we fail. Should we have a higher level check to protect against malfunctions at something like $10 or $100 USD per ETH? In such cases we'd revert to manual price setting or redeploy the contract. 
6. In `testFuzzRentFixedPrice` the values for expectedPrice and unitPrice are called by walking down the exact same codepath and then compared, so this may not be asserting anything significant. 
7. I don't believe we test the rounding behavior of `divWadUp` explicitly in our math, should we have a test for this or OK to rely on solcurity's coverage?

horsefacts:
> 1. The [Data Feeds API Reference](https://docs.chain.link/data-feeds/api-reference) say that answeredInRound is deprecated, but we use it in multiple places as a check.

Yes, you're right. We can safely remove these. Here's a [good guide](https://0xmacro.com/blog/how-to-consume-chainlink-price-feeds-safely/) on the latest recommendations.

horsefacts:
OK, previous commit should address 1, 5, 6, and 2.

src/StorageRegistry.sol
```diff
@@ -231,6 +259,16 @@ contract StorageRegistry is AccessControlEnumerable {
      */
     uint256 public priceFeedMaxAge;
 
+    /**
+     * @dev Changeable by owner.
```
horsefacts, line 263:
```suggestion
     * @dev Lower bound on acceptable price feed answer. Changeable by owner.
```

src/StorageRegistry.sol
```diff
@@ -231,6 +259,16 @@ contract StorageRegistry is AccessControlEnumerable {
...
+     */
+    uint256 public priceFeedMinAnswer;
+
+    /**
+     * @dev Changeable by owner.
```
horsefacts, line 268:
```suggestion
     * @dev Upper bound on acceptable price feed answer. Changeable by owner.
```

src/StorageRegistry.sol
```diff
@@ -505,21 +549,19 @@ contract StorageRegistry is AccessControlEnumerable {
         return (lastPriceFeedUpdateBlock == block.number) ? prevEthUsdPrice : ethUsdPrice;
     }
 
+    event Wat(uint256 ua, uint256 ts);
```
horsefacts, line 552:
```suggestion
```

PR #310, opened 2023-08-02
chore: IdRegistry + KeyRegistry review

horsefacts:
## Change Summary

Clean up comments and natspec. Use simple `expectEmit()`, which checks all indexed params and event data.

## Merge Checklist

## Additional Context

---

## PR-Codex overview
This PR focuses on improving the clarity and accuracy of code comments and function parameter names. 

### Detailed summary
- Renamed `_owner` parameter to `_initialOwner` in `TrustedCaller` and `FnameResolver` constructors.
- Updated comments in `StorageRegistry` and `KeyRegistry` functions to improve clarity.
- Renamed `owner` parameter to `_initialOwner` in `Bundler` constructor.
- Renamed `owner` parameter to `_initialOwner` in `KeyRegistry` constructor.
- Updated comments in `KeyRegistry` functions to improve clarity.

> The following files were skipped due to too many changes: `src/KeyRegistry.sol`, `src/IdRegistry.sol`


horsefacts:
Two things to consider, but I don't think either need immediate fixes.

1) The fact that FID transfers preserve the registered recovery address could be unexpected behavior for some users. This is by design, but could be a surprise if e.g. someone sells their FID to another address but the recovery settings persist. I think this is OK, since a FID doesn't really have much value separate from an associated name, but we should make this behavior clear to end users. (In fact, doing so probably discourages selling FIDs).

2) I don't like that it's possible to register an invalid or malformed key in the key registry, but there's not really a way to validate against every scheme. Again, this is more of a sharp point to warn about: consumers of the key registry should know that users can register whatever `bytes` they want as a key and it's up to the caller to validate.

varunsrin:
Good callouts, some thoughts: 

1. The original design used to reset the recovery but users may not expect this and forget to configure recovery again after they recover. Since we don't really want people to be trading fids, accepting a little badness in that case for better protection against users seems like the right tradeoff.

2. Agree, I wish we could validate but will become impractical if we want to make it extensible in the future.

varunsrin (approved):

PR #320, opened 2023-08-04
chore: add Dockerfile

horsefacts:
## Motivation

In order to test interactions between hubs and contracts, it will be helpful to have a containerized Anvil node.

## Change Summary

Add `Dockerfile` and `anvil.sh` startup script. 

## Merge Checklist

## Additional Context

Usage: 
- Ensure `.env` is populated with config values. For use in Docker, make sure values are _not_ quoted.
- Build: `docker build . -t farcaster-contracts`
- Run, passing `.env` as `--env-file`: `docker run -p 8545:8545 --env-file .env farcaster-contracts`
- Access the node at `localhost:8545`: `cast call --rpc-url http://localhost:8545 <contract address> "owner()"`. See the container logs for deployed contract addresses.

---

### Detailed summary
- Added `DEPLOYER` variable in `.env.example`
- Updated `Dockerfile` to use `ghcr.io/foundry-rs/foundry:nightly` as the base image and added build and deployment steps
- Modified `FnameResolver` contract by removing the title and notice comments
- Added `anvil.sh` script for starting the Anvil server in auto-impersonate mode
- Updated `ImmutableCreate2Deployer.sol` by modifying the `deploy` function and console logs

---

### Focus of this PR:
This PR focuses on adding Dockerfile and anvil.sh script, as well as making changes to several Solidity contracts and the .env.example file.

### Detailed summary:
- Added Dockerfile for containerization.
- Added anvil.sh script for Anvil server setup.
- Modified the `deploy` function in `ImmutableCreate2Deployer.sol` contract.
- Imported `IdRegistry` in `KeyRegistry.s.sol` contract.
- Updated variable names in `IdRegistryScript` contract.
- Updated values in the .env.example file.

---

## PR-Codex overview
This PR focuses on adding deployment functionality to the codebase. 

### Detailed summary
- Added `DEPLOYER` variable to `.env.example`
- Added Dockerfile configuration
- Added `anvil.sh` script for server setup
- Modified `deploy()` function in `ImmutableCreate2Deployer.sol` to include additional logging information


Dockerfile
```diff
@@ -0,0 +1,12 @@
+# syntax=docker/dockerfile:1.4
+
+FROM ghcr.io/foundry-rs/foundry:nightly
```
sds, line 3: Not being familiar with the general stability of this repository, I would bias towards `v1.0.0` (published 25 days ago) rather than rely on a version that is potentially broken (i.e. the CLI args we rely on for the deploy script change).

Prevents the "works on my machine" class of problems.

anvil.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -e
+
+PORT=8545
+HOST="0.0.0.0"
+SLEEP_DURATION=0.1
+
+# Start the Anvil server.
+anvil --host "$HOST" --rpc-url "$MAINNET_RPC_URL" &
```
sds, line 10: I have very little understanding of how chain forking is implemented in Anvil, but we may want to consider persisting state to a file that is mounted in a volume (`--state /var/lib/anvil/state`) so that we don't incur the cost of re-syncing that state on each startup.

anvil.sh
```diff
@@ -0,0 +1,26 @@
...
+cast rpc anvil_autoImpersonateAccount true
+
+# Deploy contracts, forking from the provided RPC URL
+forge script script/Deploy.s.sol --rpc-url "http://localhost:$PORT" --unlocked --broadcast --sender "$DEPLOYER"
+
+# Disable auto-impersonation
+cast rpc anvil_autoImpersonateAccount false
+
+# Wait for Anvil
+wait $ANVIL_PID
```
sds, line 26: `wait` is tricky. It will generally work but since `bash` is handling the initial signal and forwarding the signal to the background processes, we might not always get the behavior we want.

This may be totally fine in this context, but sometimes it can lead to confusing behavior.

A way around this is to re-exec to replace the process entirely with the desired command, so:

```bash
CMD="anvil --host $HOST --rpc-url $MAINNET_RPC_URL" --state /var/lib/anvil/state
...
$CMD & # Start anvil in the background
ANVIL_PID=$!

... (deploy the contracts) ...

kill $ANVIL_PID # Stop anvil
exec $CMD # Start anvil again, but replacing this process so it receives all signals directly instead of bash
```

Alternatively, you could create a separate container via a `docker-compose.yml` file which performs the contract deployment, keeping your base Anvil container simple (i.e. it just starts `anvil` and does nothing else).

sds (commented):
This is great! Some minor comments but overall this looks like it would be very useful.

anvil.sh
```diff
@@ -0,0 +1,26 @@
...
+ANVIL_PID=$!
+
+# Wait until Anvil is ready
+while ! nc -z localhost "$PORT"; do sleep "$SLEEP_DURATION"; done
+
+# Enable auto-impersonation
+cast rpc anvil_autoImpersonateAccount true
+
+# Deploy contracts, forking from the provided RPC URL
+forge script script/Deploy.s.sol --rpc-url "http://localhost:$PORT" --unlocked --broadcast --sender "$DEPLOYER"
```
sanjayprabhu, line 20: `DEPLOYER` here is the EOA address? What' the mnemonic that's used by default? Would be nice to be able to customize that or pass in an address that ownership will be transferred to (for things like setting the trusted caller or changing storage price)

horsefacts: Yeah, `DEPLOYER` is the EOA that will send the deployment transactions. This needs to be set as the owner of `IdRegistry` and `KeyRegistry`, and the role admin of `StorageRegistry` in order to perform the post-deploy setup. steps.

The deploy scripts assume we have the values in [`.env`](https://github.com/farcasterxyz/contracts/blob/a7321bd06bf5caae6185de9457cd10a458e0dcaf/.env.example) in the environment. I'll add some more documentation on what all these values are and maybe even reasonable defaults in ` .env.example`.

horsefacts:
Closing in favor of #321

PR #354, opened 2023-08-16
feat(KeyRegistry): validate key metadata

horsefacts:
## Motivation

We'd like to validate signer key metadata on chain when a new signer key is registered. See [FIP-7](https://github.com/farcasterxyz/protocol/discussions/103) and the comment [here](https://github.com/farcasterxyz/protocol/discussions/107#discussioncomment-6685484) on the "Application Identifiers" FIP for more details.

## Change Summary

Introduce "validator" contracts to the key registry. A validator must implement the `IMetadataValidator` interface. The `KeyRegistry` looks up the registered validator by `keyType` and `metadataType` and calls it to validate signer key metadata inside `_add`, ensuring that only keys with valid metadata can be registered onchain.

## Merge Checklist

## Additional context

Add a `KeyRegistry.integration.t.sol` test, which tests using an actual `AppIdValidator` rather than a stubbed version.

---

### Focus of the PR:
This PR focuses on adding a new interface `IMetadataValidator` and making changes related to gas usage in various test cases.

### Detailed summary:
- Added a new interface `IMetadataValidator` in `src/interfaces/IMetadataValidator.sol`.
- Updated the version pragma to `pragma solidity 0.8.21` in `src/interfaces/IMetadataValidator.sol` and `src/interfaces/IdRegistryLike.sol`.
- Added a new function `validate` in the `IMetadataValidator` interface.
- Updated gas usage in various test cases in `BundleRegistryGasUsageTest.sol`, `IdRegistryGasUsageTest.sol`, and `Bundler.gas.t.sol`.
- Updated the console log formatting in `ImmutableCreate2Deployer.sol`.
- Added a new contract `SignedKeyRequestValidatorHarness` in `test/Utils.sol`.
- Updated the imports and added a new contract `SignedKeyRequestValidator` in `test/Utils.sol`.
- Updated test cases in `SignedKeyRequestValidatorTestSuite.sol` and `KeyRegistryTestSuite.sol`.
- Updated the imports and added a new contract `SignedKeyRequestValidator` in `script/Deploy.s.sol`.
- Updated the `Bundler` contract with changes related to `keyType`, `metadataType`, and `metadata` in the `SignerParams` and `trustedRegister` functions.

> The following files were skipped due to too many changes: `src/Bundler.sol`, `test/KeyRegistry/KeyRegistry.integration.t.sol`, `test/Deploy/Deploy.t.sol`, `src/validators/SignedKeyRequestValidator.sol`, `test/validators/SignedKeyRequestValidator/SignedKeyRequestValidator.t.sol`, `test/Bundler/Bundler.t.sol`, `src/KeyRegistry.sol`, `test/KeyRegistry/KeyRegistry.t.sol`


src/KeyRegistry.sol
```diff
@@ -404,6 +422,18 @@ contract KeyRegistry is TrustedCaller, Signatures, EIP712, Nonces {
...
     //////////////////////////////////////////////////////////////*/
 
+    /**
+     * @notice Set the IdRegistry contract address. Only callable by owner.
+     *
+     * @param scheme    The key scheme associated with this validator.
+     * @param typeId    The metadata typeId associated with this validator.
+     * @param validator Contract implementing IMetadataValidator.
+     */
+    function setValidator(uint32 scheme, uint8 typeId, IMetadataValidator validator) external onlyOwner {
```
varunsrin, line 432: nit: should we allow scheme == 0 and type == 0?

src/KeyRegistry.sol
```diff
@@ -422,6 +452,16 @@ contract KeyRegistry is TrustedCaller, Signatures, EIP712, Nonces {
         KeyData storage keyData = keys[fid][key];
         if (keyData.state != KeyState.NULL) revert InvalidState();
 
+        /* First byte of metadata must encode type ID */
+        uint8 typeId = uint8(metadata[0]);
```
varunsrin, line 456: the type encoded as the first byte starts to feel like its adding complexity. should we: 

1. require the type as an additional param? 
2. merge the type with the scheme so that they are one and the same?

src/validators/AppIdValidator.sol
```diff
@@ -0,0 +1,131 @@
...
+pragma solidity 0.8.21;
+
+import {Ownable2Step} from "openzeppelin/contracts/access/Ownable2Step.sol";
+import {EIP712} from "openzeppelin/contracts/utils/cryptography/EIP712.sol";
+
+import {IMetadataValidator} from "../interfaces/IMetadataValidator.sol";
+import {IdRegistryLike} from "../interfaces/IdRegistryLike.sol";
+import {IdRegistry} from "../IdRegistry.sol";
+
+contract AppIdValidator is IMetadataValidator, Ownable2Step, EIP712 {
```
varunsrin, line 11: What we're validating is that there is a signature from the entity that requested the signer, proving that they requested the signer. It may be an application, or it may be the user themselves. 

What this really is the the `RequestFidValidator` - the name is a little clunky, but directionally more accurate about what we are validating here.

src/KeyRegistry.sol
horsefacts: Yeah, we can catch these and revert if they're zero.

src/KeyRegistry.sol
horsefacts: I dislike option 2 because it means we can't add more than one metadata type for a scheme. (Maybe an OK tradeoff in practice). 1 is probably cleaner, I'll try it.

horsefacts:
@varunsrin Updated with an explicit `metadataType` arg and the naming conventions from FIP-7. This was a big rename refactor, so I'd appreciate a review to make sure I caught everything!

src/KeyRegistry.sol
```diff
@@ -404,6 +444,20 @@ contract KeyRegistry is TrustedCaller, Signatures, EIP712, Nonces {
                                  ADMIN
     //////////////////////////////////////////////////////////////*/
 
+    /**
+     * @notice Set the IdRegistry contract address. Only callable by owner.
```
horsefacts, line 448:
```suggestion
     * @notice Set a metadata validator contract for the given keyType and metadataType. Only callable by owner.
```

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,136 @@
...
+        uint256 userFid,
+        bytes memory signerPubKey,
+        bytes calldata signedKeyRequestBytes
+    ) external view returns (bool) {
+        SignedKeyRequest memory signedKeyRequest = abi.decode(signedKeyRequestBytes, (SignedKeyRequest));
+
+        if (idRegistry.idOf(signedKeyRequest.requestSigner) != signedKeyRequest.requestingFid) return false;
+
+        /**
+         *  Safety: Since keys may only be registered once, they are not
```
horsefacts, line 99:
```suggestion
         *  Safety: Since keys may only be registered once, request signatures are not
```

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,136 @@
...
+         *  vulnerable to replay, so we omit nonce and deadline in the
+         *  validation signature.
+         */
```
horsefacts, line 102:
```suggestion
        /**
         *  Safety: Since keys may only be registered once, request
         *  signatures are not vulnerable to replay, so we omit 
         *  nonce and deadline in the validation signature.
         */
```

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,136 @@
...
+    /*//////////////////////////////////////////////////////////////
+                               VALIDATION
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @notice Validate the SignedKeyRequest metadata associated with a signer key.
+     *         (Key type 1, Metadata type 1)
+     *
+     * @param userFid               The fid of the end user adding this signer key.
+     * @param signerPubKey          The public key of the signer.
```
deodad, line 83: the name `signerPubKey` is slightly confusing since the terms signer and publicKey seem to mostly have been dropped in favor of key

horsefacts: Yeah, this the EdDSA public key that the signer is adding to the key registry. The validator checks that the metadata field is a signed ECDSA message that recovers to the owner of `requestingFid`.

horsefacts: The naming in here is pretty rough, since there's both the protocol-level signer and the metadata signature. If you have suggestions for better names in here, they are very welcome.

deodad: Yea I think since we're in the context of the `SignedKeyRequest` the prefix `signer` may be more misleading than useful since it refers to an signer from an "outer scope".

Since we use `key` instead of `pubKey` in the `add` and `addFor` methods my suggestion is just `key` so the name stays consistent. But also I'm less familiar so this might have it's own collisions.

another option, though it feels heavy, `requestedFid, requestedKey, requestingFid`

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,135 @@
...
+     *
+     *  @param requestingFid The fid of the entity requesting to add
+     *                       a signer key.
+     *  @param requestSigner Signer address. Must be the owner of
+     *                       the requestingFid fid.
+     *  @param signature     EIP-712 SignedKeyRequest signature.
+     *  @param deadline      block.timestamp after which signature expires.
+     */
+    struct SignedKeyRequest {
+        uint256 requestFid;
```
varunsrin, line 27: nit: requestingFid

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,135 @@
...
+    /*//////////////////////////////////////////////////////////////
+                                 STRUCTS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     *  @notice Signed key request specific metadata.
+     *
+     *  @param requestingFid The fid of the entity requesting to add
+     *                       a signer key.
+     *  @param requestSigner Signer address. Must be the owner of
```
varunsrin, line 21: nit: must be the owner of the requesting fid.

src/validators/SignedKeyRequestValidator.sol
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * @param key                   The EdDSA public key of the signer.
+     * @param signedKeyRequestBytes An abi-encoded SignedKeyRequest struct, provided as the
+     *                              metadata argument to KeyRegistry.add.
+     *
+     * @return true if signature is valid and signer owns requestFid, false otherwise.
+     */
+    function validate(
+        uint256, /* userFid */
+        bytes memory key,
```
varunsrin, line 92: should we also validate that the length of the key is == 32 bytes? 

this prevents people from making further mistakes when generating keys.

test/validators/SignedKeyRequestValidator/SignedKeyRequestValidator.t.sol
```diff
@@ -0,0 +1,223 @@
...
+            validator.metadataTypehash(), keccak256("SignedKeyRequest(uint256 requestFid,bytes key,uint256 deadline)")
+        );
+    }
+
+    /*//////////////////////////////////////////////////////////////
+                              VALIDATION
+    //////////////////////////////////////////////////////////////*/
+
+    function testFuzzValidate(
+        uint256 signerPk,
```
varunsrin, line 21: nit: signerPk sounds like it should be the counterpart to signerPubKey. is requesterPk more accurate here?

test/KeyRegistry/KeyRegistry.t.sol
```diff
@@ -49,140 +59,225 @@ contract KeyRegistryTest is KeyRegistryTestSuite {
...
+        address recovery,
+        uint32 keyType,
+        bytes calldata key,
+        uint8 metadataType,
+        bytes memory metadata
+    ) public {
+        keyType = uint32(bound(keyType, 1, type(uint32).max));
+        metadataType = uint8(bound(metadataType, 1, type(uint8).max));
+
+        /* We set a validator for keyType 1, type 1 during setup. Remove it.*/
```
varunsrin, line 92: we didn't call _registerValidator here, are we still actually setting the validator?

src/validators/SignedKeyRequestValidator.sol
horsefacts: I changed this away from the spec because I thought `requestFid` and `requestSigner` here were clearer and more consistent. `requestFid` is the fid of the signed key request, and `requestSigner` is the signer of the request. One field with an "-ing" and one without seemed inconsistent. Pretty minor, though, if you feel like it was clearer I'm not attached to the naming.

src/validators/SignedKeyRequestValidator.sol
horsefacts: Yeah, that's a good idea and an easy check.

src/interfaces/IMetadataValidator.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.21;
+
+interface IMetadataValidator {
+    function validate(uint256 userFid, bytes memory key, bytes memory metadata) external returns (bool);
```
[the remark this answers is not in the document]
horsefacts: We're not using it in the validator for _this specific metadata type_, but I'm not sure we want to remove it from the interface. Other, validator types might use it as part of validation. 

The idea with this interface is that the validator has access to the same args as `KeyRegistry.add`. Since there's one validator per keyType/metadataType, those are omitted here—the validator knows which key type and metadata type it's associated with implicitly.

test/KeyRegistry/KeyRegistry.t.sol
horsefacts: We are, but it's happening implicitly during `setUp`. This was an artifact of an intermediate refactor, we should use `_registerValidator` explicitly everywhere.

PR #365, opened 2023-08-21
fix: add metadata to KeyRegistry batches

horsefacts:
## Motivation

Now that adding key metadata requires a unique signature per key, we need to add metadata to the batch migration functions.

## Change Summary

Update `bulkAddKeysForMigration` add to include per-key metadata in each batch. Refactor to struct args for both `bulkAddKeysForMigration` and `bulkAddKeysForMigration`, instead of separate data arrays.

## Merge Checklist

## Additional Context

- Creating these batch structs in tests is painful, since they include arrays of items that need to be initialized an manipulated separately. I've added a couple builder libraries in `KeyRegistryTestHelpers` that make it easier to construct batches in tests.


sds (approved):
Thanks!

PR #378, opened 2023-08-28
feat: add RecoveryProxy

horsefacts:
## Motivation

We'd like to be able to change the implementation of Warpcast recovery (for example, from a Safe multisig to a custom contract) without forcing users to update their recovery addresses. Other clients may want the same capability for their own recovery implementations.

## Change Summary

Add an ownable `RecoveryProxy` contract with a single function that calls `IdRegistry.recover`. Warpcast will set this contract address as the recovery address for Warpcast-created fids. Other clients may deploy their own `RecoveryProxy` contracts to make use of this pattern. Note that this is an immutable contract, NOT an upgradeable proxy.

## Merge Checklist

---

### Detailed summary
- This PR introduces a new `RecoveryProxy` contract that allows the recovery execution logic to be changed without changing the recovery address.
- The `RecoveryProxy` contract is owned by an initial owner and can recover a fid for a user who has set the `RecoveryProxy` as their recovery address.
- The PR also includes tests for the `RecoveryProxy` contract.


src/RecoveryProxy.sol
```diff
@@ -0,0 +1,54 @@
...
+
+contract RecoveryProxy is Ownable2Step {
+    /*//////////////////////////////////////////////////////////////
+                                CONSTANTS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @dev Contract version specified using Farcaster protocol version scheme.
+     */
+    string public constant VERSION = "2023.08.23";
```
varunsrin, line 15: nit: since this is not part of the protocol, I think we can exclude this

varunsrin (approved):

src/RecoveryProxy.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.21;
+
+import {IdRegistry} from "./IdRegistry.sol";
+import {Ownable2Step} from "openzeppelin/contracts/access/Ownable2Step.sol";
+
+contract RecoveryProxy is Ownable2Step {
```
varunsrin, line 7: Might be worth adding some docs here like: 

RecoveryProxy allows the recovery execution logic to be changed without changing the recovery address. 

The proxy is set to the recovery address and it delegates permissions to execute the recovery to its owner. The owner can be changed at any time, for example from an EOA to a 2/3 multisig. This allows a recovery service operator to change the recovery mechanisms in the future without requiring each user to come online and execute a transaction.

PR #411, opened 2024-02-16
Improve Signer Removal Validation

lawyered0:
This update enhances the security and robustness of the removeSigner function in our smart contract. Previously, the function allowed for the removal of any address as a signer, without checking if the address was indeed a signer. This could lead to unnecessary state changes and events being emitted for non-signer addresses, potentially causing confusion.

To address this, we've introduced a requirement check to ensure that an address is an existing signer before it can be removed. This change prevents state modifications and event emissions for addresses that are not signers, thus tightening the contract's logic and ensuring actions reflect actual state changes.

**Key Benefits:**

Enhanced Security: By verifying that an address is a current signer before removal, we prevent unnecessary or accidental modifications to the signer list.

Improved Clarity: The contract's logic is now more straightforward, with actions closely reflecting the actual state of signers.

Reduced Confusion: Emitting events only for actual state changes makes the contract's behavior more predictable and easier to follow.

This minor yet impactful enhancement aligns with best practices in smart contract development, contributing to the overall security and maintainability of our protocol.

### DIFF
--- a/src/FnameResolver.sol
+++ b/src/FnameResolver.sol
@@ -194,6 +194,7 @@ contract FnameResolver is IExtendedResolver, EIP712, ERC165, Ownable2Step {
      * @param signer The signer address.
      */
     function removeSigner(address signer) external onlyOwner {
+	require(signers[signer], "Signer does not exist");
         signers[signer] = false;
         emit RemoveSigner(signer);
     }


talhaEth:
I think we should also check for the zero address or invalid address (address(0))

PR #490, opened 2026-08-04
feat: SnapchainConfigRegistry storage, mutators, and events

topocount:
## Motivation

Describe why this issue should be fixed and link to any relevant design docs, issues or other relevant items.

## Change Summary

Describe the changes being made in 1-2 concise sentences.

## Merge Checklist

- [ ] All [commits have been signed](https://github.com/farcasterxyz/contracts/blob/main/CONTRIBUTING.md#22-signing-commits)

## Additional Context

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---

## PR-Codex overview
This PR introduces the `ISnapchainConfigRegistry` interface and its implementation in the `SnapchainConfigRegistry` contract, establishing a governance structure for validator sets and peer lists in a Snapchain network.

### Detailed summary
- Added `ISnapchainConfigRegistry` interface with:
  - `ValidatorSet` struct.
  - Error definitions for validation.
  - Events for appending, amending, and removing validator sets.
  - Constants and storage functions.
- Implemented `SnapchainConfigRegistry` contract inheriting from `ISnapchainConfigRegistry`:
  - Defined constants for max shard IDs and public keys.
  - Storage for validator sets and peer lists.
  - Functions for managing validator sets and peer lists, including validation logic.

### DIFF
--- a/src/SnapchainConfigRegistry.sol
+++ b/src/SnapchainConfigRegistry.sol
@@ -1,0 +1,409 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.29;
+
+import {Ownable2Step} from "openzeppelin/contracts/access/Ownable2Step.sol";
+
+import {ISnapchainConfigRegistry} from "./interfaces/ISnapchainConfigRegistry.sol";
+
+/**
+ * @title SnapchainConfigRegistry
+ *
+ * @notice Canonical, owner-governed record of a Snapchain network's validator sets and gossip peer
+ *         lists. Validator-set history is append-only: earlier entries are immutable, and only the
+ *         latest may be amended or removed. Peer lists are plain settable strings.
+ *
+ * @notice See https://github.com/farcasterxyz/contracts/blob/main/docs/snapchain-config-registry.md
+ *         for the schema and the rendered-output specification.
+ *
+ * @custom:security-contact security@merklemanufactory.com
+ */
+contract SnapchainConfigRegistry is ISnapchainConfigRegistry, Ownable2Step {
+    /*//////////////////////////////////////////////////////////////
+                                CONSTANTS
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @inheritdoc ISnapchainConfigRegistry
+     */
+    string public constant VERSION = "2026.08.04";
+
+    /**
+     * @inheritdoc ISnapchainConfigRegistry
+     */
+    uint256 public constant MAX_SHARD_IDS = 16;
+
+    /**
+     * @inheritdoc ISnapchainConfigRegistry
+     */
+    uint256 public constant MAX_VALIDATOR_PUBLIC_KEYS = 128;
+
+    /**
+     * @inheritdoc ISnapchainConfigRegistry
+     */
+    uint256 public constant MAX_PEER_STRING_LENGTH = 8192;
+
+    /**
+     * @dev Allowlist of bytes permitted in a peer string, as a bitmap over 0x00-0x3F. Bit i is set
+     *      if byte i is allowed. Covers space, comma, hyphen, period, slash, the digits, and colon.
+     *
+     *      Anything a multiaddr or base58 peer id can contain is here; nothing else is. In
+     *      particular the quote and backslash are excluded, which is what makes it impossible for a
+     *      peer string to escape the TOML literal it is rendered into.
+     */
+    uint256 internal constant _PEER_CHARS_LO = 0x07FFF00100000000;
+
+    /**
+     * @dev Allowlist bitmap over 0x40-0x7F; bit i is set if byte (0x40 + i) is allowed. Covers the
+     *      uppercase letters, underscore, and the lowercase letters.
+     */
+    uint256 internal constant _PEER_CHARS_HI = 0x07FFFFFE87FFFFFE;
+
+    /*//////////////////////////////////////////////////////////////
+                                STORAGE
+    //////////////////////////////////////////////////////////////*/
+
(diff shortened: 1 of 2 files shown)


src/SnapchainConfigRegistry.sol
```diff
@@ -0,0 +1,308 @@
...
+    //////////////////////////////////////////////////////////////*/
+
+    /**
+     * @inheritdoc ISnapchainConfigRegistry
+     */
+    function appendValidatorSet(
+        uint64 effectiveAt,
+        uint32[] calldata shardIds,
+        bytes32[] calldata validatorPublicKeys
+    ) external onlyOwner {
```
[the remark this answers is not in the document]
topocount: Tests land in #492, the top of this stack — 47 of them, covering exactly the surface flagged here: duplicate/zero-key rejection, the peer-string allowlist (walked over all 256 byte values against an independently written predicate), nested-array resize on amend, and stale-key leakage after remove/re-append.

This is the known cost of reviewing a stacked series: #490 and #491 are genuinely untested in isolation, and the stack is meant to merge as a unit rather than one PR at a time. Happy to squash the four into one PR if that reads better.

src/SnapchainConfigRegistry.sol
```diff
@@ -0,0 +1,310 @@
...
+     */
+    function _writeValidatorSet(
+        uint256 index,
+        uint64 effectiveAt,
+        uint32[] calldata shardIds,
+        bytes32[] calldata validatorPublicKeys
+    ) internal {
+        _validateShardIds(shardIds);
+        _validateValidatorPublicKeys(validatorPublicKeys);
+        if (index != 0 && effectiveAt < _validatorSets[index - 1].effectiveAt) revert InvalidEffectiveAt();
```
topocount: This revert will break the config after a shard at a lower height, and completely unrelated to other shards, is entered after a shard with a higher effective_at value

topocount: Fixed — the bound is now per shard.

\`_validateEffectiveAt\` walks back from the entry being written and, for each shard it governs, compares against the most recent preceding entry governing that same shard, stopping once every shard is resolved. Entries over disjoint shards are unconstrained relative to one another, so a shard-1 rollout at height 30,000,000 now appends cleanly behind a shard-0 entry at 50,000,000.

Kept the check rather than dropping it: `get_validator_set` is an order-independent max scan so the consumer never needed ordering, but a height typed too low *for a shard it governs* backdates that set over already-committed blocks, changing which keys verify historical commits and breaking sync from genesis. Within a shard the bound stays `>=`, since sibling rollouts can land on the same height.

Spec §2.3 rewritten in #489, and the two tests that encoded the old whole-array rule used disjoint shards (0 then 1), so they were asserting exactly the broken behavior — replaced in #492 with same-shard reverts, an overlapping-shard revert, a most-recent-governing-entry case, and a disjoint-shard regression test. 51 tests pass; the golden mainnet render is unchanged.

topocount: Updating this — the fix landed, but not in the shape described above. The backward walk is gone; the bound is now held in state.

**Implementation.** `_lastEffectiveAt[shardId]` tracks the running maximum `effectiveAt` per shard, so validating an entry is a direct mapping lookup per shard rather than a scan over history. Entries governing disjoint shards are unconstrained relative to one another, so a shard-1 rollout at height 30,000,000 appends cleanly behind a shard-0 entry at 50,000,000 — the case flagged here. Within a shard the bound stays `>=`, since sibling rollouts can land on the same height.

The non-obvious part is tip mutability. `amendLatestValidatorSet` and `removeLatestValidatorSet` have to undo the tip's contribution before the bound can be re-evaluated against the entries preceding it, and a running maximum carries no history to undo from. Snapshotting only the tip's displaced value fails on repeated removals, since each removal exposes a new tip whose own displaced value is then unknown. So every entry records what it displaced in `_priorEffectiveAt[index]`, positionally aligned with its `shardIds`. All three mutators are then O(shards): append checks and records in one pass, amend rolls back and takes the append path, remove rolls back and pops. Because `appendValidatorSet` already pushes an empty entry before writing, the roll-back is a no-op on append and both paths stay identical.

**Why keep a check at all.** `get_validator_set` is an order-independent max scan, so the consumer never required ordering. But a height typed too low *for a shard the entry governs* backdates that set over already-committed blocks, changing which keys verify historical commits and breaking sync from genesis. That is worth guarding against; ordering across unrelated shards was not.

**Gas**, since the original motivation was cost: this is a worst-case win rather than a uniform one. Appends got slightly more expensive in the common case — roughly 25k gas each in the 300-entry pagination test — because snapshotting priors costs an SSTORE per shard, while the walk it replaced usually terminated in about three iterations. The gain is that the O(n) tail is gone: a shard not named by any recent entry previously cost a full-array scan, around 1.9M gas at 300 entries, and is now bounded by shard count. Storage grows about two slots per entry; the contract by 634 bytes.

**Coverage.** Spec §2.3 rewritten in #489. The two tests that encoded the old whole-array rule used disjoint shards (0 then 1), so they were asserting exactly the broken behavior — replaced in #492, alongside coverage for the roll-back: remove restores the bound, repeated removes restore it, repeated amends walk the tip back down, and an amend that retargets the tip onto different shards must undo its contribution to the shards it previously governed. Deleting the roll-back from `remove` fails three of those four, so they have teeth. 55 tests pass and the golden mainnet render is unchanged.

